PR #14, opened 2021-03-18
Fix readme formatting

Callisto13:
Because ~I~ WE totally forgot to check this in the last pr 🤦‍♀️


aclevername (approved):
LGTM

Callisto13:
Oh we don't have any auto-merge thing going on here. how do we get that?

PR #15, opened 2021-03-18
Delete all child resources when deletion occurs

aclevername:
### Description
Previously when the subscrption was deleted the child resources were left behind. This PR changes it so that the child resources get set with an `ownerReference`  pointing to the subscription. When the subscription gets deleted the [garbage collection](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/) detects the owner of these child resources has been deleted and deletes the child resources

The `testenv` doesn't handle garbage collection, so the integration tests were not updated to check for resource deletion. https://book.kubebuilder.io/reference/envtest.html#testing-considerations

### Checklist
- [X] Added tests that cover your change
- [X] Made sure the title of the PR is a good description that can go into the release notes
- [X] Added a `kind` label to the PR (e.g. `kind/feature`)

### DIFF
--- a/controllers/profile_controller.go
+++ b/controllers/profile_controller.go
@@ -19,6 +19,8 @@ package controllers
 import (
 	"context"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
 	"github.com/go-logr/logr"
 	"github.com/weaveworks/profiles/api/v1alpha1"
 	"github.com/weaveworks/profiles/pkg/git"
@@ -65,11 +67,13 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
 	pSub := v1alpha1.ProfileSubscription{}
 	err := r.Client.Get(ctx, client.ObjectKey{Name: req.Name, Namespace: req.Namespace}, &pSub)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("resource has been deleted")
+			return ctrl.Result{}, nil
+		}
 		logger.Error(err, "failed to get resource")
 		return ctrl.Result{}, err
 	}
-
-	// TODO delete if deletion timestamp != nil
 
 	// TODO replace with mutating webhook
 	if pSub.Spec.Branch == "" {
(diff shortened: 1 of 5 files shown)


tests/acceptance/acceptance_test.go
```diff
@@ -80,6 +80,16 @@ var _ = Describe("Acceptance", func() {
 			// not register as running, even though all child resources are fine
 			// :shrug_emoji:
```
Callisto13, line 81: think this comment can go since now helm release is fine

tests/acceptance/acceptance_test.go
```diff
@@ -80,6 +80,16 @@ var _ = Describe("Acceptance", func() {
...
 			Expect(podList.Items[0].Spec.Containers[0].Image).To(Equal(nginxImage))
+
+			Expect(kClient.Delete(context.Background(), &pSub)).To(Succeed())
+
+			Eventually(func() int {
+				podList = &v1.PodList{}
+				err := kClient.List(context.Background(), podList, opts...)
+				Expect(err).NotTo(HaveOccurred())
+				return len(podList.Items)
+			}, 5*time.Minute, 10*time.Second).Should(Equal(0))
```
Callisto13, line 91: could we either have this:
a) in an after each
b) in a separate test (this one talks about deploying a thing)
c) in this test which is renamed with some `By`s to make steps clear

aclevername: I prefer c)

tests/acceptance/acceptance_test.go
aclevername: :+1: I'll update the test to check the helm-release status too

controllers/profile_controller.go
```diff
@@ -65,12 +67,14 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			logger.Info("resource has been deleted")
+			return ctrl.Result{}, nil
+		}
 		logger.Error(err, "failed to get resource")
 		return ctrl.Result{}, err
 	}
 
-	// TODO delete if deletion timestamp != nil
-
 	// TODO replace with mutating webhook
```
aclevername, line 78: AFAIK since we don't use finalizers the deletion timestamp doesn't get set, the resource gets immediately removed. So deletions are handled by checking if the `get` fails because the resource does not exist.

Callisto13 (approved):
LGTM 🚀

PR #23, opened 2021-03-23
Add release actions

Callisto13:
WIP - feel free to add or amend as you like

### Description

This adds some very basic release flow automation.

I have decided to keep things super basic, therefore:
- No release-candidate nonsense
- No branching, only tags
- We don't need to release any binaries, so assets are just crds and pullable docker images (multi-arch)
- Current manual steps added to readme, can be automated further.

These are based on what the flux team do with their component releases. I have set to just push to dockerhub, but there may be other registries we want.

(TODO: set up tokens etc)


aclevername:
@Callisto13 is this ready for review or still a draft?

Callisto13:
Bit of both. I intend to come back to it after the bootstrapping mess, but you can review/alter/give thoughts etc as you want to if you have time now.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - 'v*'
+  # should be able to not have to create tag manually
+  # workflow_dispatch: {}
+
+env:
+  CONTROLLER: profiles-controller
+  ORG: weaveworks
+
+jobs:
+  # run all tests first?
```
aclevername, line 14: if its on main it should be green since it passed PR review :thinking:

Callisto13: indeed, one would hope

.github/workflows/release.yaml
```diff
@@ -0,0 +1,76 @@
...
+            org.opencontainers.image.description=${{ github.event.repository.description }}
+            org.opencontainers.image.url=${{ github.event.repository.html_url }}
+            org.opencontainers.image.revision=${{ github.sha }}
+            org.opencontainers.image.version=${{ steps.prep.outputs.VERSION }}
+            org.opencontainers.image.created=${{ steps.prep.outputs.BUILD_DATE }}
+      - name: Check images
+        run: |
+          docker buildx imagetools inspect docker.io/${{ env.ORG }}/${{ env.CONTROLLER }}:${{ steps.prep.outputs.VERSION }}
+          docker pull docker.io/${{ env.ORG }}/${{ env.CONTROLLER }}:${{ steps.prep.outputs.VERSION }}
+      - name: Generate release manifests
```
aclevername, line 65: what does this actually do? If the branch is tagged the files can't be overriden?

aclevername (approved):
lgtm for first pass, couple questions

.github/workflows/release.yaml
```diff
@@ -0,0 +1,76 @@
...
+          mkdir -p config/release
+          kustomize build ./config/crd > ./config/release/${{ env.CONTROLLER }}.crds.yaml
+          kustomize build ./config/manager > ./config/release/${{ env.CONTROLLER }}.deployment.yaml
+      - name: Do release
+        uses: softprops/action-gh-release@v1
+        with:
+          bodyPath: "docs/release_notes/${{ VERSION }}.md"
+          files: "config/release/*.yaml"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
aclevername, line 76: what does this actually do? If the branch is tagged the files can't be overriden?

Callisto13: what files?

aclevername: ohhh its artifacts, nm I'm silly

Callisto13: yeh but the job is creating a dir to hold the out files before upload to release. i don't think anything is being overwritten 🤷‍♀️ 

copied a bunch from flux

aclevername: I guess my question would be should these files be committed to the tag branch, if flux don't then who cares :shrug:

PR #28, opened 2021-03-25
Support passing values to HelmReleases

### LINKED ISSUE #3: I can customize the values of my profile with a helm values file
When deploying a Helm chart from a profile, I want to be able to provide configuration for the deployed chart.

```yaml
apiVersion: ...
kind: ProfileSubscription
spec:
   profileURL: https://github.com/testing/testing.git
   values: # This should be taken from the config for the Nginx chart and applied to the HelmRelease
      foo: value1
      bar:
      baz: value2
      oof:
      - item1
      - item2
   valuesFrom:
   - configMapKeyRef:
       # Name of the config map
       name: default-values  # mandatory
       # Namespace of the config map
       namespace: my-ns      # optional; defaults to HelmRelease namespace
       # Key in the config map to get the values from
       key: values.yaml      # optional; defaults to values.yaml
       # If set to true successful retrieval of the values file is no
       # longer mandatory
       optional: false       # optional; defaults to false

  
```

This should work in Fluxv2.

Documentation for configuring HelmReleases https://github.com/fluxcd/helm-controller/blob/main/docs/spec/v2beta1/helmreleases.md#values-overrides

This is not concerned with Secrets or ConfigMaps, if a value is exposed by the Helm chart, this can configure it.

cPu1:
### Description

Allows passing values to a `HelmRelease` by specifying the values in a `ProfileSubscriptionSpec` via the `Values` and `ValuesFrom` fields. 

Closes #3 

TODO:
- [x] Add tests

### Checklist


Callisto13:
Could you update or add an example in `examples/` to show how to use the `values` and `valuesFrom` in a profile sub?

Callisto13 (changes requested):
👍  looks good!
Just needs some tests to show values set on subscription end up somewhere.

cPu1:
> Could you update or add an example in `examples/` to show how to use the `values` and `valuesFrom` in a profile sub?

Added an example for both `values` and `valuesFrom` with a `ConfigMap`.

pkg/profile/artifact.go
```diff
@@ -68,21 +61,23 @@ func (p *Profile) createHelmRelease(ctx context.Context) error {
...
 					SourceRef: helmv2.CrossNamespaceObjectReference{
-						Kind:      gitRepositoryKind,
+						Kind:      sourcev1.GitRepositoryKind,
 						Name:      p.makeGitRepoName(),
 						Namespace: namespace,
 					},
 				},
 			},
+			Values:     p.subscription.Spec.Values,
+			ValuesFrom: p.subscription.Spec.ValuesFrom,
```
aclevername, line 80: can you update the `artifact_test.go` to test these fields are set?

cPu1: Done

Callisto13 (approved):

Issue #25, opened 2021-03-25
Show profile details from a catalog


bigkevmcd:
**What feature/behavior/change do you want?**

Show profile details

```
$ pctl show weaveworks/nginx-ingress --catalog-url foo-bar.com
Version: 1.10.1
Catalog: weaveworks (https://github.com/weaveworks/profiles)
URL: https://github.com/weaveworks/nginx-profile
Prerequisites: Kubernetes 1.18+
Description: NGINX Ingress Controller
Maintainer: WeaveWorks <gitops@weave.works>
```

**Why do you want this feature?**

Before installing a profile, I want to make sure that it will work with my cluster and that it provides what I expect.

PR #46, opened 2021-04-06
Add get endpoint to catalog url

Callisto13:
### Description

Part of https://github.com/weaveworks/profiles/issues/25.

This one needs to be merged ahead of the other side of this (https://github.com/weaveworks/pctl/pull/8), but both can be reviewed now.

### Checklist


api/v1alpha1/profilecatalogsource_types.go
```diff
@@ -31,6 +31,22 @@ type ProfileDescription struct {
 	Name string `json:"name,omitempty"`
 	// Profile description
 	Description string `json:"description,omitempty"`
+	// Version
+	// +optional
+	Version string `json:"version,omitempty"`
+	// Catalog is the name of the catalog the profile is listed in
+	// TODO necessary?
+	// +optional
+	Catalog string `json:"catalog,omitempty"`
```
bigkevmcd, line 40: We can work this out from the CatalogSource?

Callisto13: oh yeh we can get from the metadata, totally meant to do that!

pkg/catalog/export_test.go
```diff
@@ -0,0 +1,7 @@
+package catalog
+
+import "github.com/weaveworks/profiles/api/v1alpha1"
```
bigkevmcd, line 3: Not sure why this is in a test file `_test.go`?

Callisto13: Ah this is a good question.

So depending on who you ask, this is either a "valid testing pattern" or "total egregious hack". Basically by creating a `_test.go` file that is _not_ in a `_test` package, you can manipulate things like private fields as needed for testing, saving you having to make them exposed for just that purpose.

Honestly this is a convention I am still deciding whether I like or not, my gut leans towards the "egregious hack" camp, but I am happy to try it out 🤷‍♀️ .

@aclevername may also have some opinions on this.

bigkevmcd: I get how it works, and I'm trying to work out why `.profiles` can't be exported?

In any case, this feels like it's for testing state, rather than behaviour?

aclevername: Yeah this is something that they use in the standard libraries for testing, mainly to expose private fields, [example](https://golang.org/src/math/export_test.go). Its quite handy in keeping your codes public exposure, rather than making a private field public *just* for testing you can use this functionality and keep it private. We use this elsewhere in profiles as well, for example injecting the fake httpClient https://github.com/weaveworks/profiles/blob/9734639f24ecd4dbbe9bc071235a4279ee285f60/pkg/git/export_test.go#L1-L5

pkg/api/api.go
```diff
@@ -35,3 +37,15 @@ func (a *API) ProfilesHandler(w http.ResponseWriter, r *http.Request) {
 		panic(err)
 	}
 }
+
+func (a *API) ProfileHandler(w http.ResponseWriter, r *http.Request) {
+	profileName := mux.Vars(r)["name"]
+	out, err := json.Marshal(a.profileCatalog.Show(profileName))
```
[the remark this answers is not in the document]
Callisto13: oh yeh that's neater, thanks 👍

pkg/catalog/export_test.go
bigkevmcd: I don't think it needs to be exposed at all?

I think you can test the _behaviour_ ?

Callisto13: Ah right, I didn't see what you were asking. I'll look at the test again

api/v1alpha1/profilecatalogsource_types.go
Callisto13: as noted out of band - i am going to do this in a separate PR as it has revealed some weirdness with adding more than one catalog, and i want to unblock other work

README.md
```diff
@@ -10,7 +10,10 @@ Gitops native package management
 1. deploy an example catalog `kubectl apply -f examples/profile-catalog-source.yaml`
 
 1. To query the catalog API run `kubectl -n profiles-system port-forward <profiles-controller-pod-name> 8000:8000` to enable access to the API and use
-[pctl](https://github.com/weaveworks/pctl) to query, for example: `pctl search --catalog-url http://localhost:8000 <profile-name>`
+[pctl](https://github.com/weaveworks/pctl) to query, for example: `pctl --catalog-url http://localhost:8000 show <search-string>`
+
+1. To see more details on a specific Profile in the catalog, use
+[pctl](https://github.com/weaveworks/pctl): `pctl --catalog-url http://localhost:8000 show <profile-name>`
```
aclevername, line 16: same nit as in https://github.com/weaveworks/pctl/pull/8 lets use the environment variable for CATALOG_URL, makes it eaisier to read :smile:

api/v1alpha1/profilecatalogsource_types.go
Callisto13: nvm did it anyway

pkg/api/api.go
```diff
@@ -21,17 +22,24 @@ func New(profileCatalog *catalog.Catalog) API {
...
+
+func (a *API) ProfileHandler(w http.ResponseWriter, r *http.Request) {
+	profileName := mux.Vars(r)["name"]
+	marshalResponse(w, a.profileCatalog.Get(profileName))
+}
+
+func marshalResponse(w http.ResponseWriter, v interface{}) {
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(v); err != nil {
+		log.Printf("failed to encode response: %s", err)
```
aclevername, line 43: shall we change the return code if we fail to encode the response?

pkg/catalog/catalog_test.go
```diff
@@ -8,15 +8,52 @@ import (
...
+
+	Context("Get", func() {
+		BeforeEach(func() {
+			profiles := []v1alpha1.ProfileDescription{{Name: "foo"}, {Name: "bar"}, {Name: "alsofoo"}}
+			c.Add(catName, profiles...)
+		})
+
+		It("returns the requested profile", func() {
+			Expect(c.Get("foo")).To(Equal(
+				v1alpha1.ProfileDescription{Name: "foo", Catalog: catName},
```
aclevername, line 56: should this all be in 1 test with `By("")s`? It feels icky that the contexts are all cross-testing each other, e.g. if I  break `add` it also breaks `get` and `show`

Callisto13: Yeh you are right, definitely not nice 👍

pkg/api/api.go
Callisto13: 🤔  do you mean like:

```go
w.Header().Set("Content-Type", "application/json")
encoder := json.NewEncoder(w)
if err := encoder.Encode(v); err != nil {
  w.WriteHeader(500) // whatever
  if err := encoder.Encode(someErrorTypeThing); err != nil {
    log.Printf("failed to encode error setting thing: %s", err)
  }
  log.Printf("failed to encode response: %s", err)
}
```

i feel logging should be enough, not sure if failing to encode is a thing which happens easily

aclevername: I just mean adding the `w.WriteHeader(500)`. otherwise we return a `200` even though we've failed to write the response, producing a weird behaviour for the consumer of the API

Callisto13: oh so just like: 

```go
w.Header().Set("Content-Type", "application/json")
if err := json.NewEncoder(w).Encode(v); err != nil {
  w.WriteHeader(500)
  log.Printf("failed to encode response: %s", err)
}
```

aclevername: yeah

pkg/api/api_test.go
```diff
@@ -5,10 +5,11 @@ import (
 	"net/http/httptest"
 	"net/url"
 
+	"github.com/gorilla/mux"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
-	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
+	"github.com/weaveworks/profiles/api/v1alpha1"
```
aclevername, line 12: bad merge? I think we want `profilesv1` :smile:

pkg/catalog/catalog.go
```diff
@@ -3,6 +3,7 @@ package catalog
 import (
 	"strings"
 
+	"github.com/weaveworks/profiles/api/v1alpha1"
 	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
```
aclevername, line 7: duplicates

pkg/catalog/catalog.go
```diff
@@ -29,5 +33,17 @@ func (c *Catalog) Search(name string) []profilesv1.ProfileDescription {
 			profiles = append(profiles, p)
 		}
 	}
+
```
aclevername, line 36: did you want fix the `strings.Contains` to fail if `name` is empty string? Unrelated to the PR, but spotted during testing get :wink:

aclevername (approved):
LGTM, couple imports need fixing and 1 nit :smile:

pkg/api/api_test.go
Callisto13: ugh that merge was crap

pkg/catalog/catalog.go
Callisto13: should we fail or just return nothing?

Callisto13: meh will do it later

PR #47, opened 2021-04-07
Fixing the linter to show more errors and do more checks

Skarlso:
### Description

This PR will fix the linter checker to be more precise and aggressive. But still allows some wiggle room as confidence of golinter is set to 1 which ignores things like, error shouldn't be a full sentence and kClient shouldn't start with a k.

### Checklist

### Imports

Import statements now follow the following structure:

- stdlib
- thridparty
- organisational third party
- package local

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -15,7 +15,7 @@ type API struct {
 	profileCatalog *catalog.Catalog
 }
 
-// New creates a new mux based api router.
+// New returns a new mux based api router.
 func New(profileCatalog *catalog.Catalog) API {
 	r := mux.NewRouter()
 	a := API{
@@ -27,7 +27,7 @@ func New(profileCatalog *catalog.Catalog) API {
 	return a
 }
 
-// ProfilesHandler is the handler for /profiles events.
+// ProfilesHandler is the handler for /profiles requests.
 func (a *API) ProfilesHandler(w http.ResponseWriter, r *http.Request) {
 	profileName := r.URL.Query().Get("name")
 	out, err := json.Marshal(a.profileCatalog.Search(profileName))
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -5,14 +5,17 @@ import (
 	"net/http"
 
 	"github.com/gorilla/mux"
+
 	"github.com/weaveworks/profiles/pkg/catalog"
 )
 
+// API defines a catalog router.
 type API struct {
 	*mux.Router
 	profileCatalog *catalog.Catalog
 }
 
+// New creates a new mux based api router.
 func New(profileCatalog *catalog.Catalog) API {
 	r := mux.NewRouter()
 	a := API{
@@ -24,6 +27,7 @@ func New(profileCatalog *catalog.Catalog) API {
 	return a
 }
 
+// ProfilesHandler is the handler for /profiles events.
 func (a *API) ProfilesHandler(w http.ResponseWriter, r *http.Request) {
 	profileName := r.URL.Query().Get("name")
 	out, err := json.Marshal(a.profileCatalog.Search(profileName))
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -3,27 +3,27 @@ package catalog
 import (
 	"strings"
 
-	"github.com/weaveworks/profiles/api/v1alpha1"
+	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
 )
 
-// Catalog contains profiles which were gathered from the underlying cluster.
+// Catalog provides an in-memory cache of profiles from the cluster which can be queried easily.
 type Catalog struct {
-	profiles []v1alpha1.ProfileDescription
+	profiles []profilesv1.ProfileDescription
 }
 
 // New creates a new, empty catalog.
 func New() *Catalog {
-	return &Catalog{profiles: []v1alpha1.ProfileDescription{}}
+	return &Catalog{profiles: []profilesv1.ProfileDescription{}}
 }
 
 // Add adds p profiles to the catalog.
-func (c *Catalog) Add(p ...v1alpha1.ProfileDescription) {
+func (c *Catalog) Add(p ...profilesv1.ProfileDescription) {
 	c.profiles = append(c.profiles, p...)
 }
 
-// Search returns catalogs which contain the given `name`.
-func (c *Catalog) Search(name string) []v1alpha1.ProfileDescription {
-	var profiles []v1alpha1.ProfileDescription
+// Search returns profile descriptions that contain `name` in their names.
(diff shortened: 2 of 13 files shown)


pkg/api/api.go
```diff
@@ -5,14 +5,17 @@ import (
...
 	"github.com/weaveworks/profiles/pkg/catalog"
 )
 
+// API defines a catalog router.
 type API struct {
 	*mux.Router
 	profileCatalog *catalog.Catalog
 }
 
+// New creates a new mux based api router.
```
bigkevmcd, line 18: The usual styling for these is "returns" rather than "creates".

pkg/api/api.go
```diff
@@ -24,6 +27,7 @@ func New(profileCatalog *catalog.Catalog) API {
 	return a
 }
 
+// ProfilesHandler is the handler for /profiles events.
```
bigkevmcd, line 30: How about "requests" rather than "events" here?

pkg/catalog/catalog.go
```diff
@@ -6,18 +6,22 @@ import (
...
 func New() *Catalog {
 	return &Catalog{profiles: []v1alpha1.ProfileDescription{}}
 }
 
+// Add adds p profiles to the catalog.
 func (c *Catalog) Add(p ...v1alpha1.ProfileDescription) {
 	c.profiles = append(c.profiles, p...)
 }
 
+// Search returns catalogs which contain the given `name`.
```
bigkevmcd, line 24: This doesn't seem quite right, the return type is a slice of `ProfileDescription` ?

pkg/catalog/catalog.go
```diff
@@ -6,18 +6,22 @@ import (
 	"github.com/weaveworks/profiles/api/v1alpha1"
 )
 
+// Catalog contains profiles which were gathered from the underlying cluster.
```
bigkevmcd, line 9: I think this might be better represented as...

```Catalog provides an in-memory cache of profiles from the cluster which can be queried easily."

What do you think?

File: pkg/git/git.go [hunk 5]
```diff
@@ -8,17 +8,21 @@ import (
...
 	"github.com/go-logr/logr"
-	"github.com/weaveworks/profiles/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/util/yaml"
+
+	"github.com/weaveworks/profiles/api/v1alpha1"
 )
 
+// HTTPClient defines an http client which then can be used to test the
+// handler code.
 //go:generate counterfeiter -o fakes/fake_http_client.go . HTTPClient
```
bigkevmcd, line 18: I don't think this is quite right?

pkg/git/git.go
```diff
@@ -8,17 +8,21 @@ import (
...
 type HTTPClient interface {
 	Get(string) (*http.Response, error)
 }
 
 var httpClient HTTPClient = http.DefaultClient
 
+// GetProfileDefinition returns a definition based on a url and a branch.
 func GetProfileDefinition(repoURL, branch string, log logr.Logger) (v1alpha1.ProfileDefinition, error) {
```
bigkevmcd, line 26: How about we change this to return a `*v1alpha1.ProfileDefinition` ?

This simplifies the return types, and also clarifies the ownership of the value?

pkg/profile/profile.go
```diff
@@ -2,8 +2,9 @@ package profile
 
 import (
 	"github.com/go-logr/logr"
-	"github.com/weaveworks/profiles/api/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/weaveworks/profiles/api/v1alpha1"
```
bigkevmcd, line 7: I tend to name these something like `profilesv1alpha1` rather than leaving the default import `v1alpha1`.

bigkevmcd (commented):
Nothing too major, some wee tidyups.

pkg/git/git.go
Skarlso: Isn't it? I think this interface is solely there to be able to mock it out, right?

Skarlso: Hmm, maybe I didn't quite wrote this correctly.

pkg/catalog/catalog.go
Skarlso: Ops, right you are.

pkg/catalog/catalog.go
Skarlso: I think you are missing something from this comment. :) @bigkevmcd

pkg/git/git.go
Skarlso: While I usually agree, here, I would like to avoid refactoring, because this PR only deals with the Linter.

That being said, I don't see why this should share up the stack? What would be the benefit if having this as a pointer? The related calls elsewhere, do use a value and then pass that value to other functions which do not require a pointer.

Could you elaborate your view a bit further please?

Skarlso: Further, if this would be a pointer, definition in `Profile` would have to be a pointer and would thus require nil checks all over the place for valid values and property access.

pkg/git/git.go
Skarlso: How about:

`// HTTPClient defines a testable base HTTPClient that can be mocked in tests.`

pkg/catalog/catalog.go
bigkevmcd: Hah...a typo in my markup meant that GitHub didn't render it :-)

Skarlso: yeah, that sounds good. :)

pkg/catalog/catalog_test.go
```diff
@@ -3,20 +3,21 @@ package catalog_test
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/weaveworks/profiles/api/v1alpha1"
+
+	profilesv1alpha1 "github.com/weaveworks/profiles/api/v1alpha1"
```
aclevername, line 7: can we call this `profilev1` similar to ho we have `helmv2` & `sourcev1`?

Skarlso: Yeah that sounds good to me. It's shorter to. :) Are you okay with that @bigkevmcd ?

pkg/git/git.go
aclevername: I think the fact its used for testing doesn't really belong in the comment description, its more just 
`// HTTPClient defines an interface for HTTP get requests` or something like that

aclevername (commented):
apart from 1 issue left open to resolves it lgtm :smile:

pkg/git/git.go
Skarlso: Fair point! :)

aclevername (approved):

PR #48, opened 2021-04-07
Update profile subscription resource status to show artifact resource state

aclevername:
### Description

This is a first pass on having the profile subscription status be a resemblance of the state of the artifact resources. For now if any of the artifact resources report `Ready != True` the message on the condition is bubbled up to the profile subscription status.

Example output during a creation:

```
k get profilesubscriptions.weave.works nginx-profile-test --watch-only
NAME                 READY   STATUS                        AGE
nginx-profile-test   Unknown   reconciliation in progress    0s
nginx-profile-test   False     HelmChart 'default/default-nginx-profile-test-nginx-nginx-server' is not ready   0s
nginx-profile-test   Unknown   Reconciliation in progress                                                       0s
nginx-profile-test   True      all artifact resouces ready                                                      8s
```

If I manually break the `helmrelease` resource the status changes to:
```
k get profilesubscriptions.weave.works nginx-profile-test
NAME                 READY   STATUS                                                                           AGE
nginx-profile-test   False   HelmChart 'default/default-nginx-profile-test-nginx-nginx-server' is not ready   94s
```
### Checklist


controllers/profile_controller.go
```diff
@@ -79,35 +94,56 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+	}
+	return ctrl.Result{}, nil
 }
 
-func (r *ProfileSubscriptionReconciler) patchStatusRunning(ctx context.Context, pSub *v1alpha1.ProfileSubscription, logger logr.Logger) {
-	pSub.Status.State = "running"
-	pSub.Status.Message = ""
-	r.patchStatus(ctx, pSub, logger)
-}
+func (r *ProfileSubscriptionReconciler) patchStatus(ctx context.Context, pSub *v1alpha1.ProfileSubscription, logger logr.Logger, readyStatus, reason, message string) {
```
Skarlso, line 136: I'm wondering if this shouldn't return the possible errors it generates. Most of the places this is called from also bubble up errors anyways. WDYT?

controllers/profile_controller.go
```diff
@@ -79,35 +94,56 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			return ctrl.Result{}, err
+		}
+	}
 
-func (r *ProfileSubscriptionReconciler) patchStatusFailing(ctx context.Context, pSub *v1alpha1.ProfileSubscription, logger logr.Logger, message string) {
-	pSub.Status.State = "failing"
-	pSub.Status.Message = message
-	r.patchStatus(ctx, pSub, logger)
+	if len(artifactStatus.NotReadyConditions) == 0 {
+		r.patchStatus(ctx, &pSub, logger, readyTrue, "ArtifactsReady", "all artifact resouces ready")
```
Skarlso, line 120: Making this return an error though would make this code part a bit different.

pkg/profile/artifact.go
```diff
@@ -7,54 +7,180 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	p.log.Info("creating GitRepository", "resource", gitRes.ObjectMeta.Name)
+	if err := p.client.Create(ctx, gitRes); err != nil {
 		return fmt.Errorf("failed to create GitRepository resource: %w", err)
 	}
 
-	switch kind := p.definition.Spec.Artifacts[0].Kind; kind {
```
Skarlso, line 25: Interesting. I just read that we will support multiple ways for artifacts. Things like, raw yaml for Kubernetes resources and such. Or is that not this part of the code? :) I might be mistaken. I'm just wondering what prompted this refactor.

aclevername: this bit is just moving the existing code, it still exists in this file. Currently we only support 1 artifact, that can be of Kind Kustomize (which includes raw yaml) or Helm.

controllers/profile_controller.go
aclevername: yeah its an interesting one, here if the patch fails we are in a weird state, the previous "actions" have succeeded, but we were not able to notify the resource it worked. Should this cause a failure? I think probably but its not 100% clear cut.

I will update the code to handle the error outside of the

aclevername: done

pkg/profile/artifact.go
Skarlso: Gotcha, thanks.

pkg/profile/artifact.go
```diff
@@ -7,54 +7,180 @@ import (
...
+	return gitResExists && helmResExists && kustomizeResExists, gitRes, helmRes, kustomizeRes, nil
+}
+
+func getNotReadyCondition(conditions []metav1.Condition) metav1.Condition {
+	for _, condition := range conditions {
+		if condition.Type == "Ready" && string(condition.Status) != "True" {
+			return condition
+		}
+	}
+	return metav1.Condition{}
```
Skarlso, line 125: What do you think about instead of comparing to an empty struct, return a `ready` boolean? You do that in other places, I think it would be nice and consistent to do that here too?

pkg/profile/artifact.go
```diff
@@ -7,54 +7,180 @@ import (
...
+	return Status{
+		ResourcesExist:     resourcesExist,
+		NotReadyConditions: p.checkResourcesReady(gitRes, helmRes, kustomizeRes),
+	}, nil
+}
+
+func (p *Profile) checkResourcesReady(gitRes *sourcev1.GitRepository, helmRes *helmv2.HelmRelease, kustomizeRes *kustomizev1.Kustomization) []metav1.Condition {
+	var notReadyConditions []metav1.Condition
+	if gitRes != nil {
+		if condition := getNotReadyCondition(gitRes.Status.Conditions); condition != (metav1.Condition{}) {
```
Skarlso, line 69: The change to getNotReady would then result in this:

```go
		if ok, condition := getNotReadyCondition(gitRes.Status.Conditions); ok {
			notReadyConditions = append(notReadyConditions, condition)
		}
```
which reads nice, I think :)

pkg/profile/artifact.go
aclevername: I did have that originally, but it refactored it out as it looked ugly haha, I'm quite torn on which looks better. I sort of think this approach is equiavlent to returning `error` and comparing `err!=nil`. WDYT?

Skarlso: Heh, yeah. I can see that. :)
I usually avoid the structure compare because it works only as long as all the fields are comparable in the struct and then if you add another field which isn't, like a slice, then this part of the code will break. Which is fine I guess, but then you'd have to update the code to also compare slices and slice values and that's just a drag really. :)

So in the long term it's usually more convenient to just return a yay or nay if possible.

I'm not trying to convince you, I just wanted to tell you my reasoning behind it. :)

config/crd/bases/weave.works_profilesubscriptions.yaml
```diff
@@ -95,13 +105,76 @@ spec:
...
+                    reason:
+                      description: reason contains a programmatic identifier indicating
+                        the reason for the condition's last transition. Producers
+                        of specific condition types may define expected values and
+                        meanings for this field, and whether the values are considered
+                        a guaranteed API. The value should be a CamelCase string.
+                        This field may not be empty.
+                      maxLength: 1024
+                      minLength: 1
+                      pattern: ^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$
```
Skarlso, line 152: >  The value should be CamelCase string.

Shouldn't this be enforced by this pattern? Or is that should not a hard rule?

Skarlso: Also, nit: CamelCase is actually PascalCase. CamelCase is camelCase. :)

aclevername: this file is auto generated by the operator-sdk :smiley:

Skarlso: OH BAH! I was wondering about that. But wasn't sure. :D :D Guess I'll drop over and create an issue hahaha.

pkg/profile/artifact.go
aclevername: mhmm thats fair, in this case we are relying on a pretty standard struct that shouldn't realistically ever be changing in format (this is a core part of kubernetes api), I'm inclined to leave it as it is

config/crd/bases/weave.works_profilesubscriptions.yaml
aclevername: hahaha :smile:

pkg/profile/artifact.go
Skarlso: Sure, I'm fine with that. :) 👍

pkg/profile/artifact.go
```diff
@@ -7,54 +7,180 @@ import (
...
+		if err != nil {
+			return false, nil, nil, nil, err
+		}
+	}
+	return gitResExists && helmResExists && kustomizeResExists, gitRes, helmRes, kustomizeRes, nil
+}
+
+func getNotReadyCondition(conditions []metav1.Condition) metav1.Condition {
+	for _, condition := range conditions {
+		if condition.Type == "Ready" && string(condition.Status) != "True" {
```
bigkevmcd, line 121: https://pkg.go.dev/github.com/rmb938/hostport-allocator/api/meta#IsStatusConditionTrue ?

controllers/profile_controller.go
```diff
@@ -79,43 +94,74 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			messages = append(messages, condition.Message)
+			if string(condition.Status) == readyFalse {
+				status = readyFalse
+			}
+		}
+		if err := r.patchStatus(ctx, &pSub, logger, status, "ArtifactNotReady", strings.Join(messages, ",")); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
```
Skarlso, line 141: WDYT about dropping the else and replacing it with a return in the len check if? That way the flow of the function is not broken.

Skarlso (approved):
LGTM in my yet limited structural and architectural understanding of the task and code :)

pkg/profile/artifact.go
aclevername: oo nice! I'll refactor to use :smile:

controllers/profile_controller.go
aclevername: sounds good!

aclevername:
@Callisto13 if your going to rename the PR you could at least review it :wink: :laughing:

pkg/profile/artifact.go
```diff
@@ -7,54 +7,181 @@ import (
...
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to get resource: %w", err)
+	}
+
+	return true, nil
+}
+
 // MakeArtifacts creates and returns a slice of runtime.Object values, which if
 // applied to a cluster would deploy the profile as a HelmRelease.
```
Callisto13, line 141: is this true?

pkg/profile/artifact.go
```diff
@@ -7,54 +7,181 @@ import (
...
-	case "HelmChart":
-		if err := p.createHelmRelease(ctx); err != nil {
+	if helmRes != nil {
+		p.log.Info("creating HelmRelease", "resource", helmRes.ObjectMeta.Name)
+		if err := p.client.Create(ctx, helmRes); err != nil {
 			return fmt.Errorf("failed to create HelmRelease resource: %w", err)
 		}
-	case "Kustomize":
-		if err := p.createKustomization(ctx); err != nil {
+	} else {
```
Callisto13, line 44: meh not really a fan of `else`s

pkg/profile/artifact.go
```diff
@@ -7,54 +7,181 @@ import (
...
+		}
+	}
+
+	if kustomizeRes != nil {
+		kustomizeResExists, err = p.getResourceIfExists(ctx, kustomizeRes)
+		if err != nil {
+			return false, nil, nil, nil, err
+		}
+	}
+	return gitResExists && helmResExists && kustomizeResExists, gitRes, helmRes, kustomizeRes, nil
```
Callisto13, line 117: this is one hell of a return signature 😂

pkg/profile/artifact.go
```diff
@@ -7,54 +7,181 @@ import (
...
+		return nil, nil, nil, fmt.Errorf("failed to create GitRepository resource: %w", err)
 	}
 
-	objs = append(objs, gr, hr)
-	return objs, nil
+	var helmRes *helmv2.HelmRelease
+	var kustomizeRes *kustomizev1.Kustomization
+
+	switch kind := p.definition.Spec.Artifacts[0].Kind; kind {
+	case "HelmChart":
```
Callisto13, line 169: do we want to const or type these kinds?

pkg/profile/artifact.go
bigkevmcd: This code looks complex, I suspect we could simplify it with [`%T`](https://golang.org/pkg/fmt/)

controllers/profile_controller.go
```diff
@@ -55,7 +67,10 @@ func (r *ProfileSubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&profilesv1.ProfileSubscription{}, builder.WithPredicates(
 			predicate.GenerationChangedPredicate{},
-		)).
+		)). // Owns ensures that changes to resouces owned by the pSub cause the pSub to get requeued
+		Owns(&sourcev1.GitRepository{}).
+		Owns(&helmv2.HelmRelease{}).
+		Owns(&kustomizev1.Kustomization{}).
```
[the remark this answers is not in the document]
aclevername: Not sure, but it's pretty sweat!

pkg/profile/artifact.go
aclevername: I'll update

pkg/profile/artifact.go
aclevername: Temporary, once multiple artifacts is added this will change

pkg/profile/artifact.go
aclevername: It's that or multiple returns for each Boolean :/

pkg/profile/artifact.go
aclevername: That's being done as part of @Skarlso PR 😃

pkg/profile/artifact.go
bigkevmcd: Do you need the booleans?

I think you can simplify this by returning nil if the object isn't found? (which you're doing).

aclevername: I think its just simpler to keep the logic around if the resources existed inside the func. If no boolean is returned and just nil objects, how do we know `kustomizeRes` is nil because there isn't a kustomize artifact in the profile definition or because it was missing in the cluster?

aclevername: In the multiple artifacts PR I've changed it to this https://github.com/weaveworks/profiles/pull/61/files#diff-b425bde9d0ac4c1a499461b4ef45b4d9983833ab9025dccb91e85ac893df3a45R93-R118
```
	for _, helmRes := range helmResources {
		helmResExists, err := p.getResourceIfExists(ctx, helmRes)
		if err != nil || !helmResExists {
			return false, nil, nil, nil, err
		}
	}

	for _, kustomizeRes := range kustomizeResources {
		kustomizeResExists, err := p.getResourceIfExists(ctx, kustomizeRes)
		if err != nil || !kustomizeResExists {
			return false, nil, nil, nil, err
		}
	}
	return true, gitRes, helmResources, kustomizeResources, nil

Issue #50, opened 2021-04-08
Support Helm Chart Repository Chart installations


bigkevmcd:
**What feature/behavior/change do you want?**

Currently, we support charts that are embedded in the profile itself, but we should be able to support upstream charts from chart repositories.

**Why do you want this feature?**

To simplify the installation of a provided Helm Chart.

This should use a `HelmRepository` rather than a `GitRepository`.

Issue #26, opened 2021-03-25
Install by generating a ProfileSubscription


bigkevmcd:
**What feature/behavior/change do you want?**

Install a profile by generating a ProfileSubscription, that can install Profile with a secret or configmap. (assuming config map exists in cluster.)

```
$ pctl install weaveworks/nginx-ingress --namespace europe --config-secret=service-config-secret
apiVersion: weave.works/v1alpha1
kind: ProfileSubscription
...
```

I can then apply that yaml to my cluster and see that my profile is installed with my values set.

**Why do you want this feature?**

I want to be able to create a subscription to a specific profile.

Callisto13:
Questions about this feature:

1. We talk about "subscribing" to a profile, so what language do we want here: `pctl subscribe` or `pctl install`
2. Do we want subscription tied to catalogs (at this point in the implementation)? ie: if you ask to install a profile which is _not_ in your catalog, do we want to error?

Callisto13:
blocked on https://github.com/weaveworks/profiles/issues/25

Callisto13:
@bigkevmcd were you going to summarise the chat we had about this last week here?

Callisto13:
For initial implementation we are going to only install (or in the case of this story generate a yaml for installation) a profile which is available in a catalog.

```
pctl --catalog-url 127.0.0.7:8000 install nginx
```

In the future we will add the following features:
1. Install profile _not_ in catalog from raw url
2. Allow orgs to "lock down" clusters so that users can _only_ install profiles which are listed in catalogs

but these can be left for now (i will create follow-u stories)

bigkevmcd:
I think this should write out a profile.yaml, subscription, and whatever resources are generated by the profile installation (HelmRelease, GitRepository etc.).

Callisto13:
Bumping up to M points as we need to think about how we do ownership of children

Callisto13:
@bigkevmcd could you explain why we have decided to keep all generated artifacts in the instance repo? We have some concerns that it breaks a lot of the purpose of pctl/profiles

bigkevmcd:
I'll dig further, but a Profile is nested Profiles.

Callisto13:
From the discussion today we are no longer generating anything other than a profile subscription yaml when we install.

Callisto13:
@Skarlso see comment above

PR #54, opened 2021-04-10
Add ability to access helm charts through a provided Helm URL

Skarlso:
### Description

Added ability to get helm charts not only from the profile's local repository, which copies the thing in whole
but also with a defined helm chart url which points to a chart like https://org.github.io/charts.

### Checklist

### Issue

#50

### Ownership Change

We introduced a small change in the generating of the Artifacts.

Now, there are two methods.

- MakeArtifacts
This will generate artifacts with ownership set to Controller.
- MakeOwnerlessArtifacts
This will generate artifacts without any ownership set.

The later will be used in https://github.com/weaveworks/profiles/issues/26 by install to generate a set of artifacts which then will generate yaml files which can then be installed. This is so that we don't duplicate artifacts generation in pctl and just reuse what's already there.


controllers/profile_controller_test.go
```diff
@@ -151,7 +154,7 @@ var _ = Describe("ProfileController", func() {
 				subscriptionName := "git-resource-already-exists-error"
 				profileURL := nginxProfileURL
 
-				gitRefName := fmt.Sprintf("%s-%s-%s", subscriptionName, "nginx-profile", "main")
+				gitRefName := fmt.Sprintf("%s-%s-%s", subscriptionName, "nginx-profile", "support-helm-urls")
```
Skarlso, line 157: TODO: We need to change / deal with this. This is currently pinning to a repository here: https://github.com/weaveworks/nginx-profile/pull/6 which need to change the chart value to HelmChartLocal.

api/v1alpha1/profile_types.go
```diff
@@ -28,17 +37,23 @@ type ProfileDefinitionSpec struct {
...
 
+// Artifact defines a bundled resource of the components for this profile.
 type Artifact struct {
 	// Name is the name of the Artifact
 	Name string `json:"name,omitempty"`
 	// Path is the local path to the Artifact in the Profile repo
 	Path string `json:"path,omitempty"`
-	// Kind is the kind of artifact: HelmChart or Kustomize
+	// Kind is the kind of artifact: HelmChartLocal or Kustomize
 	Kind string `json:"kind,omitempty"`
```
bigkevmcd, line 50: We could add a kubebuilder validation annotation for the enumeration here?

api/v1alpha1/profile_types.go
```diff
@@ -28,17 +37,23 @@ type ProfileDefinitionSpec struct {
...
+	// HelmURL is the URL of the Helm repository containing a Helm chart and possible values
```
bigkevmcd, line 51: I wonder if this isn't extractable into a separate struct?

The tuple of (`chart repository`, `chart`, `version`) feels like a value type.

pkg/profile/artifact.go
```diff
@@ -6,28 +6,51 @@ import (
...
+	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
 )
 
+// location defines where the helm chart is located.
+// remote -> in a helm repository
+// local -> in the profile repository
+type location int
+
+var local location = 0
+var remote location = 1
```
bigkevmcd, line 25: These could be `iota`ed consts?

pkg/profile/artifact.go
```diff
@@ -6,28 +6,51 @@ import (
...
+
 // CreateArtifacts creates and inserts objects to the cluster to deploy the
 // profile as a HelmRelease.
 func (p *Profile) CreateArtifacts(ctx context.Context) error {
-	if err := p.createGitRepository(ctx); err != nil {
-		return fmt.Errorf("failed to create GitRepository resource: %w", err)
-	}
-
+	// TODO don't depend on 0.
```
bigkevmcd, line 30: We could expand this to `Artifacts[0]` ?

pkg/profile/artifact.go
```diff
@@ -6,28 +6,51 @@ import (
...
+		}
+		if err := p.createHelmRelease(ctx, local); err != nil {
 			return fmt.Errorf("failed to create HelmRelease resource: %w", err)
 		}
-	case "Kustomize":
+	case profilesv1.HelmChartRemoteKind:
+		if p.definition.Spec.Artifacts[0].HelmURL == "" {
+			return fmt.Errorf("helmURL for helm repository for remote helm artifact must not be empty")
+		}
+		//TODO: figure out what to do with the chart name and version. Maybe they could be inferred?
```
bigkevmcd, line 43: Can you expand a bit more on this? can't you use `HelmChartVersion` ?

pkg/profile/artifact_test.go
```diff
@@ -172,9 +172,82 @@ var _ = Describe("Profile", func() {
...
+				p = profile.New(pDef, pSub, fakeClient, logr.Discard())
+				err := p.CreateArtifacts(ctx)
+				Expect(err).NotTo(HaveOccurred())
+
+				gitRefName := fmt.Sprintf("%s-%s-%s", subscriptionName, "repo-name", branch)
+				helmRepository := sourcev1.HelmRepository{}
+				err = fakeClient.Get(ctx, client.ObjectKey{Name: gitRefName, Namespace: namespace}, &helmRepository)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(helmRepository.Spec.URL).To(Equal("https://org.github.io/charts"))
+				Expect(helmRepository.OwnerReferences).To(HaveLen(1))
```
bigkevmcd, line 209: I wonder if we couldn't extract this to an `IsOwnedBy(...)` matcher?

pkg/profile/artifact.go
Skarlso: I asked in this Slack. It's more like a conversation about how deep we want to surface Helm Chart values in profile yaml data. At some point it ought to be too much.

pkg/profile/artifact.go
Skarlso: For sure 👍

api/v1alpha1/profile_types.go
Skarlso: Good idea. The separation would be nice indeed.

api/v1alpha1/profile_types.go
Skarlso: Absolutely 👍

pkg/profile/artifact.go
Skarlso: I'm not sure I follow, sorry. The `[0]` will go away. Jake has a branch somewhere where he's fixing the reliance on it. We have to enumerate the artifacts anyways. So this is just a note to not use this at all.

api/v1alpha1/profile_types.go
Skarlso: This is done.

controllers/profile_controller_test.go
```diff
@@ -36,10 +36,173 @@ var _ = Describe("ProfileController", func() {
 		Expect(k8sClient.Create(context.Background(), &nsp)).To(Succeed())
 	})
 
-	Context("Create", func() {
+	//Context("Create Local", func() {
```
Skarlso, line 39: @Skarlso TODO: Uncomment this to see if things works once the multiple artifacts thing is handled.

controllers/profile_controller_test.go
```diff
@@ -180,7 +343,8 @@ var _ = Describe("ProfileController", func() {
 						Namespace: namespace,
 					},
 					Spec: profilesv1.ProfileSubscriptionSpec{
-						ProfileURL: profileURL,
+						ProfileURL: nginxProfileURL,
+						Branch:     "support-helm-urls",
```
Skarlso, line 347: TODO: Remove once multiple artifacts are handled.

api/v1alpha1/profile_types.go
Skarlso: Added.

bigkevmcd: I was thinking more of something like:
```
// +kubebuilder:validation:Enum=HelmChartLocal;Kustomize
```

Which would embed this into the CRD (and validate).

But this also hints at a possible smell `HelmChartLocal` and `Kustomize` don't feel at the same level of abstraction?

Skarlso: I think you are looking at an outdated revision. I think this is github messing up the resolved comment.
I completely removed the notion of `Local` and now I'm just using the optional Chart value and Path value. :)

pkg/profile/artifact.go
aclevername: https://github.com/weaveworks/profiles/pull/61

aclevername:
@Skarlso is this ready for review? I see a lot of TODO comments :smile:

Skarlso:
@aclevername As discussed, no. :D I have to merge your branch into it and that will require some changes. :)

pkg/profile/artifact.go
```diff
@@ -156,36 +177,50 @@ func (p *Profile) MakeArtifacts() ([]runtime.Object, error) {
...
+				return nil, nil, nil, nil, fmt.Errorf("failed to create HelmRelease resource: %w", err)
 
 			}
 			helmResources = append(helmResources, helmRes)
-		case "Kustomize":
+			if artifact.Path != "" {
+				gitRes, err = p.makeGitRepository()
+				if err != nil {
+					return nil, nil, nil, nil, fmt.Errorf("failed to create GitRepository resource: %w", err)
+				}
```
aclevername, line 199:
```suggestion
			if artifact.Path != "" && gitRes == nil {
				gitRes, err = p.makeGitRepository()
				if err != nil {
					return nil, nil, nil, nil, fmt.Errorf("failed to create GitRepository resource: %w", err)
				}
```

No need to keep recreating this each time :smile:

Skarlso: Ah, good point! Thanks :)

pkg/profile/artifact_test.go
```diff
@@ -27,12 +27,16 @@ import (
 const (
 	subscriptionName     = "mySub"
 	namespace            = "default"
-	branch               = "main"
+	branch               = "support-helm-urls"
```
Skarlso, line 30: TODO: Don't forget to change this to main once https://github.com/weaveworks/nginx-profile/pull/6 is merged.

pkg/profile/artifact_test.go
```diff
@@ -423,7 +451,7 @@ var _ = Describe("Profile", func() {
 				Expect(profilesv1.AddToScheme(scheme)).To(Succeed())
 				p = profile.New(pDef, pSub, fakeClient, logr.Discard())
 				err := p.CreateArtifacts(ctx)
-				Expect(err).To(MatchError(ContainSubstring("failed to create GitRepository resource")))
+				Expect(err).To(MatchError(ContainSubstring("failed to create resource: no kind is registered for the type v1beta1.GitRepository in scheme")))
```
Skarlso, line 454: @aclevername Curious. This suddenly is throwing a different error. I'll investigate why but it's not obvious at first glance. I changed too much. :D

Skarlso: Oh I think I know actually.

Skarlso: Yeah, I think this is sort of fine? Now the log is more precise and tells you that the kind doesn't exist. I can add the Kind that it was trying to create into the Error as well. That helps too.

Skarlso: Now it will say:
```
failed to create v1beta1.GitRepository
```
instead of
```
failed to create resource
```

@aclevername Is that okay?

aclevername: I think its worth mentioning which resource error'd? Helps for debugging

Skarlso: Yep added that. Now this looks like this:

```
failed to create HelmRelease: no kind is registered for the type v2beta1.HelmRelease
```

pkg/profile/artifact.go
```diff
@@ -140,52 +133,63 @@ func (p *Profile) getResourceIfExists(ctx context.Context, res client.Object) (b
...
+				if err != nil {
+					return nil, fmt.Errorf("failed to create GitRepository resource: %w", err)
+				}
+			} else if artifact.Chart != nil {
+				helmRep, err := p.makeHelmRepository(artifact.Chart.HelmURL, artifact.Chart.HelmChart)
+				if err != nil {
+					return nil, fmt.Errorf("failed to create HelmRepository resource: %w", err)
+				}
+				objs = append(objs, helmRep)
+			}
```
aclevername, line 167: shall we add some tests + validation checks for ensuring `chart` and `path` are not both set?

Skarlso: Definitely. I'm actually writing that just now. :) And increasing coverage as well.

bigkevmcd: It'd be nice to see this done via a `Validate()` method on the ProfileDefinition?

Something like...

```
import "knative.dev/pkg/apis"

(p ProfileDefinition) Validate() *apis.Fielderror {
...
}

It doesn't _have_ to be the knative apis.FieldError...but it does have nice functionality for indicating which fields within the resource are invalid.

https://pkg.go.dev/knative.dev/pkg/apis#FieldError

Comment (Skarlso) [reply pkg/profile/artifact.go:167]: I like how we are thinking similarly. :) The only difference is that I actually put the Validate on the Artifact. Are you okay with that @bigkevmcd?

Comment (bigkevmcd) [reply pkg/profile/artifact.go:167]: Makes sense for it to be at the level of the thing being validated :-)

Comment (bigkevmcd) [reply pkg/profile/artifact.go:170]: For comparison, the output of using the standard errors for this:

```go
func (in Artifact) Validate() error {
    if in.Chart != nil && in.Path != "" {
      return apis.ErrMultipleOneOf("chart", "path")
    }
    return nil
 }
```

Would get you

`validation failed on artifact helmChartArtifactName1: expected exactly one, got both: chart, path"`
`validation failed on artifact helmChartArtifactName1: chart and path cannot both be defined"`

Skarlso: Added error. But I'm wrapping it out, because I like the context of the chart name, if that is okay.

controllers/profile_controller_test.go
```diff
@@ -36,10 +37,10 @@ var _ = Describe("ProfileController", func() {
 		Expect(k8sClient.Create(context.Background(), &nsp)).To(Succeed())
 	})
 
-	Context("Create", func() {
+	Context("Create with multiple artifacts", func() {
 		DescribeTable("Applying a Profile creates the correct resources", func(pSubSpec profilesv1.ProfileSubscriptionSpec) {
 			subscriptionName := "foo"
-			branch := "main"
+			branch := "support-helm-urls"
```
aclevername, line 43: TODO replace

pkg/profile/artifact.go
```diff
@@ -140,52 +133,66 @@ func (p *Profile) getResourceIfExists(ctx context.Context, res client.Object) (b
...
```
aclevername, line 170: can this just be `else` rather than `else if`? Its sort of awkward, because `Validation` passed we know that its always going to work with an `if else`, but its sort of leaking that knowledge into this pkg. https://github.com/weaveworks/eksctl/issues/3566. 

Perhaps the `Artifact` struct should live here, rather than in the `profile_types` package? Just throwing random thought out.

pkg/profile/artifact.go
```diff
@@ -212,7 +219,40 @@ func (p *Profile) makeGitRepository() (*sourcev1.GitRepository, error) {
...
+		},
+		TypeMeta: metav1.TypeMeta{
+			Kind:       sourcev1.HelmRepositoryKind,
+			APIVersion: sourcev1.GroupVersion.String(),
+		},
+		Spec: sourcev1.HelmRepositorySpec{
+			URL: url,
+		},
+	}
+	err := controllerutil.SetControllerReference(&p.subscription, helmRepo, p.client.Scheme())
```
aclevername, line 236: now that we are wrapping all the resources in the `runtime.Object` can we do this in 1 place for all objects rather than repeating in each `makeResX`?

pkg/profile/artifact_test.go
```diff
@@ -147,7 +162,8 @@ var _ = Describe("Profile", func() {
 		When("the artifact exists", func() {
 			var (
 				gitRes                       *sourcev1.GitRepository
-				helmRes1, helmRes2           *helmv2.HelmRelease
+				helmRep                      *sourcev1.HelmRepository
+				helmRes1, helmRes2, helmRes3 *helmv2.HelmRelease
```
aclevername, line 166: minor nit: I think we can mdoify `helmRes2` instead of adding a new `helmRes3`. Two exist just for testing that it supported multiple `HelmChart` kinds, so the 2nd can be a remote chart :smile:

pkg/profile/artifact.go
Skarlso: What do you think about just using ifs? Since the validation passed we know that only ONE should ever be there anyways.

pkg/profile/artifact_test.go
Skarlso: Yeah, sure, no problem. The BBD test cases are a bit confusing to be honest. :D It's not readily apparent which variables are safe to be re-used.

aclevername (commented):
Missing unit tests for `CreateArtifacts`

pkg/profile/artifact_test.go
aclevername: happy to pair to go over if if you want :smile:

pkg/profile/artifact.go
Skarlso: Sure, yeah, we can do that. :)

controllers/profile_controller_test.go
```diff
@@ -132,18 +168,20 @@ var _ = Describe("ProfileController", func() {
...
 					profile.Status.Conditions[0].Type == "Ready" &&
-					profile.Status.Conditions[0].Status == metav1.ConditionStatus("True") &&
-					profile.Status.Conditions[0].Message == "all artifact resouces ready"
+					profile.Status.Conditions[0].Status == "True" &&
+					profile.Status.Conditions[0].Message == "all artifact resources ready"
 			}, 10*time.Second).Should(BeTrue())
 		},
 			Entry("a single Helm chart with no supplied values", profilesv1.ProfileSubscriptionSpec{
 				ProfileURL: nginxProfileURL,
+				Branch:     "support-helm-urls",
```
aclevername, line 177: TODO replace

pkg/profile/artifact.go
```diff
@@ -140,52 +133,90 @@ func (p *Profile) getResourceIfExists(ctx context.Context, res client.Object) (b
 // MakeArtifacts creates and returns a slice of runtime.Object values, which if
 // applied to a cluster would deploy the profile as a HelmRelease.
 func (p *Profile) MakeArtifacts() ([]runtime.Object, error) {
```
aclevername, line 135: is this func redundant now? We can just have the private one public

aclevername (approved):
lgtm, couple todos left

pkg/profile/artifact.go
Skarlso: @aclevername I don't know what else might call this. Is there anything outside profile which might call this in any ways? I didn't dare deleting it.

Callisto13:
Could you add some explanation for the whole "Ownerless" thing into the PR description? That context will be useful in future.

Skarlso:
@Callisto13 Added description. Does it make sense? If you think we should add more detail, I can do that.

Issue #63, opened 2021-04-13
Remove associated catalog when a ProfileCatalogSource is removed


cPu1:
When a `ProfileCatalogSource` is removed via the HTTP API, its associated catalog continues to linger in the in-memory cache. The controller should handle this case and remove the catalog.

Issue #52, opened 2021-04-08
I can update a catalog in my cluster


Callisto13:
The API needs to handle altering profiles in the array which are associated with that catalog

Issue #51, opened 2021-04-08
I can delete a catalog source from my cluster


Callisto13:
The API needs to handle removing things from the profiles array which are associated with that catalog.

This would be useful to have sooner rather than later, because right now the integration tests keep failing when run multiple times as the in memory list persists between runs

cPu1:
>I can delete a catalog from my cluster

>The API needs to handle removing things from the profiles array which are associated with that catalog.

Is this issue about removing a profile from a catalog, or a catalog itself?

Callisto13:
a catalog itself. there is another story to handle updating a catalog (adding/removing profiles from an existing catalog)

PR #65, opened 2021-04-13
Keep in-memory cache synchronized with ProfileCatalogSource

cPu1:
### Description

Closes https://github.com/weaveworks/profiles/issues/63
Closes https://github.com/weaveworks/profiles/issues/51
Closes https://github.com/weaveworks/profiles/issues/52

The cache is currently built with the assumption that there can't be multiple catalogs with the same name in different namespaces, so the catalog's name itself is used as the key. 
Related: https://github.com/weaveworks/profiles/issues/66

### Checklist


pkg/catalog/catalog.go
```diff
@@ -7,38 +7,44 @@ import (
...
+func (c Catalog) Update(catalogName string, profiles ...profilesv1.ProfileDescription) {
+	for i := range profiles {
+		profiles[i].Catalog = catalogName
 	}
+	c[catalogName] = profiles
+}
+
+// Remove removes the specified catalog.
+func (c Catalog) Remove(catalogName string) {
+	delete(c, catalogName)
```
Callisto13, line 27:
```suggestion
// Remove removes the specified catalog source.
func (c Catalog) Remove(sourceName string) {
	delete(c, sourceName)
```

Just to make sure nobody gets confused

pkg/catalog/catalog_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 var _ = Describe("Catalog", func() {
```
Callisto13, line 11: TODO: add test for `Remove`

pkg/catalog/catalog.go
cPu1: The rest of the code uses `catalogName` and I'm not sure if `sourceName` is clearer than that.

bigkevmcd: How about just "name" ?

It's `Catalog`.Remove()` what do catalogs remove? sources? what could the name be? the name of a source to remove?

cPu1: I'm not sure anyone new to the codebase would be able to deduce that `catalog.Delete(name string)` deletes a source. Nowhere in the catalog code we've used the word source. Also, this changelist did not introduce the word `catalogName`, it's already being used in other catalog operations.

Callisto13: > this changelist did not introduce the word catalogName, it's already being used in other catalog operations

True, and that is a mistake given we have already conflated catalogs and sources multiple times in discussion.

cPu1: I'm happy to rename all `catalogName` references to `sourceName`. WDYT?

pkg/catalog/catalog_test.go
cPu1: Done

tests/acceptance/acceptance_test.go
```diff
@@ -257,19 +257,61 @@ var _ = Describe("Acceptance", func() {
...
+			})
+		})
+
+		Context("delete", func() {
+			It("clears the in-memory cache when a ProfileCatalogSource is deleted", func() {
+				Expect(kClient.Delete(context.Background(), &pCatalog)).To(Succeed())
+				Eventually(func() error {
+					_, err := getProfile(profileName)
+					return err
+				}).Should(MatchError(ContainSubstring("got 404")))
```
aclevername, line 313: minor nit, but is it worth going a `Get` first, to prove its there before deleting? Atm this test could pass if the setup for the test is incorrect

pkg/catalog/catalog.go
Callisto13: yeh i think that would help unmuddle things 😅 , thanks!

tests/acceptance/acceptance_test.go
cPu1: Done.

tests/acceptance/acceptance_test.go
```diff
@@ -251,25 +251,71 @@ var _ = Describe("Acceptance", func() {
...
 						Name:        "nginx-2",
 						Description: "nginx 1",
-						Catalog:     catalogName,
+						Catalog:     sourceName,
 					},
 				))
 			})
 		})
 
+		getProfile := func(profileName string) (profilesv1.ProfileDescription, error) {
```
bigkevmcd, line 260: Given that this is only used in a test, it doesn't need to return an error, it can just `Expect` things and fail.

bigkevmcd: Also, it could return a pointer to a ProfileDescription?

cPu1: >Given that this is only used in a test, it doesn't need to return an error, it can just Expect things and fail.

There are multiple callers of this function, one of which expects a specific type of error to have occurred,, so moving the assertions to this function will not let us do that.

bigkevmcd: You could pass that in as a param?

cPu1: There are multiple code paths in this function that can return an error. Which of those calls will that param be checked against? The `(value, error)` return type is a well-established pattern. Moreover, it's more verbose to replace `return err` calls with assertions.

Callisto13:
Gonna get this one finished and merged while Chetan is on holiday 👍

pkg/api/api_test.go
```diff
@@ -77,5 +118,25 @@ var _ = Describe("Api", func() {
 				Expect(rr.Body.String()).To(ContainSubstring(`{"name":"nginx-1","description":"nginx 1","catalog":"catalog"}`))
 			})
 		})
+
+		When("the requested profile does not exist", func() {
+			BeforeEach(func() {
+				fakeCatalog.GetReturns(nil)
+			})
+
+			It("returns the profile summary from the catalog", func() {
```
aclevername, line 127:
```suggestion
			It("returns 404", func() {
```

pkg/api/api_test.go
```diff
@@ -77,5 +118,25 @@ var _ = Describe("Api", func() {
...
+				req, err := http.NewRequest("GET", "/profiles", nil)
+				req = mux.SetURLVars(req, map[string]string{"catalog": sourceName, "profile": profileName})
+				Expect(err).NotTo(HaveOccurred())
+
+				rr := httptest.NewRecorder()
+				handler := http.HandlerFunc(catalogAPI.ProfileHandler)
+
+				handler.ServeHTTP(rr, req)
+
+				// Check the status code is what we expect.
```
aclevername, line 137:
```suggestion
				// Check the status code is what we expect.
```

aclevername (approved):

Issue #69, opened 2021-04-20
Prevent arbitrary artifact resources modifications


aclevername:
**What feature/behavior/change do you want?**
If I create a profile that creates a `HelmRelease` resource, if a users tries to manually edit the spec the controller should overwrite any changes and it bring it back to the desired state. Essentially mirroring the same behaviour that would occur if you tried to modify a pod owned by a deployment.

The same should occur if a user deletes the `HelmRelease`, it should get recreated.

**Why do you want this feature?**
To prevent the artifact resources from drifting out of sync

PR #75, opened 2021-04-21
reconcile artifact resources

aclevername:
### Description
https://github.com/weaveworks/profiles/issues/69

### Checklist

### DIFF
--- a/controllers/profile_controller.go
+++ b/controllers/profile_controller.go
@@ -103,10 +103,9 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
-	instance := profile.New(pDef, pSub, r.Client, logger, ctx)
+	instance := profile.New(ctx, pDef, pSub, r.Client, logger)
 
-	err = instance.ReconcileArtifacts()
-	if err != nil {
+	if err = instance.ReconcileArtifacts(); err != nil {
 		if err := r.patchStatus(ctx, &pSub, logger, readyFalse, "CreateFailed", "error when reconciling profile artifacts"); err != nil {
 			return ctrl.Result{}, err
 		}
--- a/controllers/profile_controller.go
+++ b/controllers/profile_controller.go
@@ -72,6 +72,7 @@ func (r *ProfileSubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error
 		)). // Owns ensures that changes to resources owned by the pSub cause the pSub to get requeued
 		Owns(&sourcev1.GitRepository{}).
 		Owns(&helmv2.HelmRelease{}).
+		Owns(&sourcev1.HelmRepository{}).
 		Owns(&kustomizev1.Kustomization{}).
 		Complete(r)
 }
@@ -103,26 +104,21 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
 	}
 
 	instance := profile.New(pDef, pSub, r.Client, logger)
+
+	err = instance.ReconcileArtifacts(ctx)
+	if err != nil {
+		if err := r.patchStatus(ctx, &pSub, logger, readyFalse, "CreateFailed", "error when reconciling profile artifacts"); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, err
+	}
+
 	artifactStatus, err := instance.ArtifactStatus(ctx)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	if !artifactStatus.ResourcesExist {
-		err = instance.CreateArtifacts(ctx)
-		if err != nil {
-			if err := r.patchStatus(ctx, &pSub, logger, readyFalse, "CreateFailed", "error when creating profile artifacts"); err != nil {
-				return ctrl.Result{}, err
-			}
-			return ctrl.Result{}, err
-		}
-		artifactStatus, err = instance.ArtifactStatus(ctx)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-	}
-
-	if len(artifactStatus.NotReadyConditions) == 0 {
+	if artifactStatus.ResourcesExist && len(artifactStatus.NotReadyConditions) == 0 {
 		return ctrl.Result{}, r.patchStatus(ctx, &pSub, logger, readyTrue, "ArtifactsReady", "all artifact resources ready")
 	}
 
--- a/controllers/profile_controller.go
+++ b/controllers/profile_controller.go
@@ -103,9 +103,9 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
-	instance := profile.New(pDef, pSub, r.Client, logger)
+	instance := profile.New(pDef, pSub, r.Client, logger, ctx)
 
-	err = instance.ReconcileArtifacts(ctx)
+	err = instance.ReconcileArtifacts()
 	if err != nil {
 		if err := r.patchStatus(ctx, &pSub, logger, readyFalse, "CreateFailed", "error when reconciling profile artifacts"); err != nil {
 			return ctrl.Result{}, err
@@ -113,7 +113,7 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
-	artifactStatus, err := instance.ArtifactStatus(ctx)
+	artifactStatus, err := instance.ArtifactStatus()
 	if err != nil {
 		return ctrl.Result{}, err
 	}
--- a/pkg/profile/artifact.go
+++ b/pkg/profile/artifact.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"time"
 
 	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
@@ -24,9 +23,8 @@ type Status struct {
 	NotReadyConditions []metav1.Condition
 }
 
-// CreateArtifacts generate and creates the objects in the cluster to deploy the
-// profile
-func (p *Profile) CreateArtifacts(ctx context.Context) error {
+// ReconcileArtifacts ensures the artifact resources are applied to the cluster
+func (p *Profile) ReconcileArtifacts(ctx context.Context) error {
 	objs, err := p.MakeArtifacts()
 	if err != nil {
 		return err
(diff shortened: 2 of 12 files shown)


controllers/profile_controller.go
```diff
@@ -72,6 +72,7 @@ func (r *ProfileSubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error
 		)). // Owns ensures that changes to resources owned by the pSub cause the pSub to get requeued
 		Owns(&sourcev1.GitRepository{}).
 		Owns(&helmv2.HelmRelease{}).
+		Owns(&sourcev1.HelmRepository{}).
```
aclevername, line 75: think this was missing from the introduction of helmrepo resources

pkg/profile/artifact.go
```diff
@@ -37,19 +34,63 @@ func (p *Profile) CreateArtifacts(ctx context.Context) error {
...
+	}
+}
+
+func (p *Profile) updateResource(existingObj, desiredObj client.Object) error {
+	switch desiredObj := desiredObj.(type) {
+	case *sourcev1.GitRepository:
+		if !GitRepoRequiresUpdate(existingObj.(*sourcev1.GitRepository), desiredObj) {
+			return nil
+		}
+		existingObj.(*sourcev1.GitRepository).Spec = desiredObj.Spec
```
aclevername, line 68: anybody know if there is a way to not have to type converter twice :vomiting_face: ?

controllers/profile_controller.go
Skarlso: Ah yes, nice.

controllers/profile_controller.go
```diff
@@ -102,27 +103,22 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
-	instance := profile.New(pDef, pSub, r.Client, logger)
-	artifactStatus, err := instance.ArtifactStatus(ctx)
+	instance := profile.New(ctx, pDef, pSub, r.Client, logger)
+
+	err = instance.ReconcileArtifacts()
```
Skarlso, line 108: Would you mind inlining this in the iff below with `if err := ..; err != nil`? :)

pkg/profile/artifact.go
```diff
@@ -37,19 +34,62 @@ func (p *Profile) CreateArtifacts(ctx context.Context) error {
...
+		if err := p.reconcileArtifact(obj); err != nil {
+			return err
 		}
 	}
 
 	p.log.Info("all artifacts created")
 	return nil
 }
+func (p *Profile) reconcileArtifact(desiredObj client.Object) error {
+	existingObj := desiredObj.DeepCopyObject().(client.Object)
```
Skarlso, line 46: What's the point of the deepclone here if you are immediately getting it, and Object key is used from the desiredObj?

pkg/profile/artifact.go
```diff
@@ -37,19 +34,62 @@ func (p *Profile) CreateArtifacts(ctx context.Context) error {
...
+	err := p.client.Get(p.ctx, client.ObjectKeyFromObject(desiredObj), existingObj)
+	if err == nil {
+		return p.updateResource(existingObj, desiredObj)
+	} else if apierrors.IsNotFound(err) {
+		p.log.Info("creating...", "kind", desiredObj.GetObjectKind().GroupVersionKind().Kind, "resource", desiredObj.GetName())
+		if err := p.client.Create(p.ctx, desiredObj); err != nil {
+			return fmt.Errorf("failed to create %s: %w", desiredObj.GetObjectKind().GroupVersionKind().Kind, err)
+		}
+		return nil
+	} else {
```
Skarlso, line 56: The else is not needed. Just return.

pkg/profile/artifact.go
```diff
@@ -37,19 +34,62 @@ func (p *Profile) CreateArtifacts(ctx context.Context) error {
...
 
 	p.log.Info("all artifacts created")
 	return nil
 }
+func (p *Profile) reconcileArtifact(desiredObj client.Object) error {
+	existingObj := desiredObj.DeepCopyObject().(client.Object)
+	err := p.client.Get(p.ctx, client.ObjectKeyFromObject(desiredObj), existingObj)
+	if err == nil {
+		return p.updateResource(existingObj, desiredObj)
+	} else if apierrors.IsNotFound(err) {
```
bigkevmcd, line 50: You don't need the `else` here, you've returned if `err == nil`.

pkg/profile/artifact.go
bigkevmcd, line 56: Same here...you've returned

pkg/profile/artifact.go
```diff
@@ -37,19 +34,62 @@ func (p *Profile) CreateArtifacts(ctx context.Context) error {
...
+		return nil
+	} else {
+		return fmt.Errorf("failed to get resource %s %s/%s: %w", desiredObj.GetObjectKind().GroupVersionKind().Kind, desiredObj.GetNamespace(), desiredObj.GetName(), err)
+	}
+}
+
+func (p *Profile) updateResource(existingObj, desiredObj client.Object) error {
+	switch desiredObj := desiredObj.(type) {
+	case *sourcev1.GitRepository:
+		if !GitRepoRequiresUpdate(existingObj.(*sourcev1.GitRepository), desiredObj) {
```
bigkevmcd, line 64: You can pass in `desiredObj` here instead of using another type assertion (same for below).

pkg/profile/artifact.go
```diff
@@ -37,19 +34,62 @@ func (p *Profile) CreateArtifacts(ctx context.Context) error {
...
+			return nil
+		}
+		existingObj.(*sourcev1.GitRepository).Spec = desiredObj.Spec
+	case *sourcev1.HelmRepository:
+		if !HelmRepoRequiresUpdate(existingObj.(*sourcev1.HelmRepository), desiredObj) {
```
bigkevmcd, line 69: It feels like `HelmRepoRequiresUpdate` is only exported for the tests, it's not used outside of this package?

pkg/profile/git_repo.go
```diff
@@ -0,0 +1,46 @@
...
+	case existingRes.Spec.URL != newRes.Spec.URL:
+		return true
+	case existingRes.Spec.Reference.Branch != newRes.Spec.Reference.Branch:
+		return true
+	default:
+		return false
+	}
+}
+
+func (p *Profile) makeGitRepository() (*sourcev1.GitRepository, error) {
```
bigkevmcd, line 22: This never returns an `error`?

pkg/profile/helm.go
```diff
@@ -0,0 +1,112 @@
...
+	return helmRepo, nil
+}
+
+func (p *Profile) makeHelmRepoName(name string) string {
+	repoParts := strings.Split(p.subscription.Spec.ProfileURL, "/")
+	repoName := repoParts[len(repoParts)-1]
+	return join(p.subscription.Name, repoName, p.subscription.Spec.Branch, name)
+}
+
+func (p *Profile) makeHelmRelease(artifact profilesv1.Artifact) (*helmv2.HelmRelease, error) {
```
bigkevmcd, line 64: This never returns an error?

pkg/profile/kustomize.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta1"
+	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// KustomizeRequiresUpdate checks if the git kustomization resource requires updating
+func KustomizeRequiresUpdate(existingRes, desiredRes *kustomizev1.Kustomization) bool {
```
bigkevmcd, line 14: Could this use `reflect.DeepEqual(desireRes.Spec, existingRes.Spec)`?

bigkevmcd (changes requested):

pkg/profile/artifact.go
aclevername: yes, that is true. I don't like testing private functions, but at the same time I felt it was important to have tests for the update checking. Not sure what to do tbh

pkg/profile/git_repo.go
aclevername: this was already here, just moved it. Will update

pkg/profile/helm.go
aclevername: this was already here, just moved it. Will update

pkg/profile/artifact.go
aclevername: could you expand? I am passing `desiredObj` in, so I don't qutie follow

Skarlso (commented):
Most of my comments I see Kevin already added, I have one more question though.

pkg/profile/kustomize.go
aclevername: good question. I leaned towards not doing `DeepEqual`  where there are fields in the struct that we don't specifically set. These fields might be getting defaulted or set by a webhook. Essentially we only check for updates to fields we specifically set.

pkg/profile/artifact.go
bigkevmcd: If it was me, I wouldn't put those tests in the `_test` package, that's used for testing the external interface of a package, and you're testing the internal behaviour here?

Skarlso: What's wrong with testing private functions? Unit tests can test whatever they want, can't they? I mean, I see that we mostly test out of package with `_test`, but I feel like exporting this function for the sake of testing is also a smell in itself. I think. :)

pkg/profile/artifact.go
aclevername: I need an object of the correct type for resource get read into

Skarlso: But you are converting it into client.Object.

Skarlso: So, I don't know, maybe I'm getting something wrong here, but, if you just create a new client.Object and `Get` that, get will provide all the type and information that it further requires. It's not like, you update a field or something afterwards. So I don't think the DeepCopy in this case adds value? everything will be overwritten anyways?

Skarlso: And and also, shouldn't you pass in `&existingObj` instead?

Skarlso: Ignore me. It's an interface. 🤦

Skarlso (approved):
LGTM

pkg/profile/artifact.go
aclevername: > If it was me, I wouldn't put those tests in the _test package, that's used for testing the external interface of a package, and you're testing the internal behaviour here?

I'm not sure where to put them then :thinking: I really dislike testing private functions, but having these tests are important (while the checking logic checks individual fields, if we can do some other `deepequal` its less of a problem). The only other alternative I see is testing all of these cases through the public function they are used in, which means bloating out the `ReconcileArtifact` tests even more.

We could make the func private, and then use the `export_test.go` pattern to export is publicly for the tests only, but its still the same problem of testing a private func, just not making this func public for no reason.

bigkevmcd: I don't think we need the `export_test` pattern either, just move the tests into the package, you're not testing the external behaviour.

pkg/profile/git_repo.go
```diff
@@ -0,0 +1,45 @@
...
+// GitRepoRequiresUpdate checks if the git repository resource requires updating
+func GitRepoRequiresUpdate(existingRes, newRes *sourcev1.GitRepository) bool {
+	switch {
+	case existingRes.Spec.URL != newRes.Spec.URL:
+		return true
+	case existingRes.Spec.Reference.Branch != newRes.Spec.Reference.Branch:
+		return true
+	default:
+		return false
+	}
```
cPu1, line 19:
```suggestion
	return existingRes.Spec.URL != newRes.Spec.URL || existingRes.Spec.Reference.Branch != newRes.Spec.Reference.Branch
```

pkg/profile/helm.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// HelmRepoRequiresUpdate checks if the helm repository resource requires updating
+func HelmRepoRequiresUpdate(existingRes, newRes *sourcev1.HelmRepository) bool {
+	switch {
+	case existingRes.Spec.URL != newRes.Spec.URL:
+		return true
+	default:
+		return false
+	}
```
cPu1, line 38:
```suggestion
	return existingRes.Spec.URL != newRes.Spec.URL:
```

pkg/profile/git_repo.go
aclevername: I think this makes it harder to read. The select is more lines, but much more readable

pkg/profile/helm.go
```diff
@@ -0,0 +1,111 @@
...
+			URL: url,
+		},
+	}
+}
+
+func (p *Profile) makeHelmRepoName(name string) string {
+	repoParts := strings.Split(p.subscription.Spec.ProfileURL, "/")
+	repoName := repoParts[len(repoParts)-1]
+	return join(p.subscription.Name, repoName, p.subscription.Spec.Branch, name)
+}
```
cPu1, line 61: This is identical to `makeGitRepoName`. This can be renamed to `makeRepoName` and reused for both Git and Helm repos.

pkg/profile/helm.go
aclevername: This is fair, I'd be inclined to stick with the select since thats the pattern used across the update checkers, but I could be convinced here... :shrug:

pkg/profile/helm.go
aclevername: I just moved it as part of this PR, but yeah thats a fair refactor I'll add in

Skarlso: The difference is that it adds the `name` at the join. And it looks awkward to call `makeRepoName("")`. But it could be two functions but a single function being called inside those two functions.

Skarlso: Actually that wouldn't work. It would end up being `asdf-asdf-` with an extra `-` at the end.

Skarlso: If you use this function for two elements they will have the same name which will cause a name collision.

Skarlso: Oh! But you need it. Because if you have a helmchart which also generates the same resource, and you don't add the name which is the Chart's name then the name will be the same for the same resource.

PR #85, opened 2021-04-26
return 404 in catalog when no results found

aclevername:
### Checklist

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -37,12 +37,12 @@ func (c *Catalog) Search(name string) []profilesv1.ProfileDescription {
 }
 
 // Get returns the profile description `profileName`.
-func (c *Catalog) Get(catalogName, profileName string) profilesv1.ProfileDescription {
+func (c *Catalog) Get(catalogName, profileName string) (bool, profilesv1.ProfileDescription) {
 	for _, p := range c.profiles {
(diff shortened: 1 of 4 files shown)


pkg/catalog/catalog.go
```diff
@@ -37,12 +37,12 @@ func (c *Catalog) Search(name string) []profilesv1.ProfileDescription {
 }
 
 // Get returns the profile description `profileName`.
-func (c *Catalog) Get(catalogName, profileName string) profilesv1.ProfileDescription {
+func (c *Catalog) Get(catalogName, profileName string) (bool, profilesv1.ProfileDescription) {
 	for _, p := range c.profiles {
```
bigkevmcd, line 41: how about returning `*profilesv1.ProfileDescription` which means you can return `nil` if it doesn't exist?

aclevername: I knew you would point this out! :laughing: I didn't want to return a pointer, because then you can modify whats in the inmemory cache?

aclevername:
turns out this is being done as part of https://github.com/weaveworks/profiles/pull/65

pkg/catalog/catalog.go
bigkevmcd: that's really an implementation detail of the fact that it's an inmemory cache?

You could do a DeepCopy on it, and return a pointer to that?

PR #98, opened 2021-05-03
Fix manifest namespaces and align release process with flux

Skarlso:
### Description

The alignment is by using a single prepared yaml file with all data attached. I've also fixed the namespace in the manifest files so they can be used out of the box.

### Checklist


config/manager/kustomization.yaml
```diff
@@ -13,5 +13,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: localhost:5000/profiles-controller
-  newTag: latest
+  newName: skarlso/profiles-controller
```
aclevername, line 16: 🤔

Skarlso: What the heck... I swear I committed the fix to this at least twice. :D Something must be off from rebasing.

Skarlso: Thanks for spotting it. I had to do a lot of testing yesterday on my own fork until this all worked. :)

config/prepare/manager_config_patch.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: system
 spec:
```
aclevername, line 5: does it need the namespace to do the patch?

Skarlso: Well done, another good spot! :)

config/manager/manager.yaml
```diff
@@ -1,15 +1,8 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  labels:
-    control-plane: controller-manager
-  name: system
----
```
aclevername, line 7: since the namespace is now defined outside of the `manager` dir does that cause any problems?

aclevername (approved):
lgtm

config/manager/manager.yaml
Skarlso: Like what?

aclevername: no idea 😄 but I noticed this is done elsewhere, so I don't think it matters

Skarlso: :D Nah, it's good, because the main prepare manifest creates it. The only thing could be if it's applied without a kustomize I guess, but that's the approach flux uses as well.

PR #99, opened 2021-05-04
Remove the image editing in the Kustomization file

Skarlso:
### Description

The image editing is anyways done in the Makefile. The Kustomization file should not edit that setting.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from a cluster
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
 deploy: manifests kustomize ## Deploy controller in the configured Kubernetes cluster in ~/.kube/config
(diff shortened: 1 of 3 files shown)


bigkevmcd (approved):

aclevername (commented):
do we need to change how `make local-env` works now this points to the remote by default?

Skarlso:
> do we need to change how `make local-env` works now this points to the remote by default?

local-env already does this:

```
cd config/manager && $(KUSTOMIZE) edit set image controller=localhost:5000/${IMG}
```

Skarlso:
Oh snap, wait. I thought that will work without this.. But of course it doesn't. Now that's a bummer.

Skarlso:
Ah, it does work. Because it then builds it and so applies the changes which were made in place. But... I had to edit the `image` tag to be the one we are setting now up! So there was a bug in there....

aclevername (approved):

PR #103, opened 2021-05-05
prevent recursive profile definitions

aclevername:
### Description

Previously we only checked 1 deep, now we check for any occurrences of a recursive profile artifact

### Checklist

### DIFF
--- a/pkg/profile/artifact.go
+++ b/pkg/profile/artifact.go
@@ -173,6 +173,14 @@ func (p *Profile) getResourceIfExists(res client.Object) (bool, error) {
 // MakeOwnerlessArtifacts generates artifacts without owners for manual applying to
 // a personal cluster.
 func (p *Profile) MakeOwnerlessArtifacts() ([]runtime.Object, error) {
+	return p.makeOwnerlessArtifacts([]string{p.profileRepo()})
+}
+
+func (p *Profile) profileRepo() string {
+	return p.subscription.Spec.ProfileURL + ":" + p.subscription.Spec.Branch
+}
+
+func (p *Profile) makeOwnerlessArtifacts(profileRepos []string) ([]runtime.Object, error) {
 	var (
 		objs   []runtime.Object
 		gitRes *sourcev1.GitRepository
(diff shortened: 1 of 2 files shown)


pkg/profile/artifact.go
```diff
@@ -184,17 +192,21 @@ func (p *Profile) MakeOwnerlessArtifacts() ([]runtime.Object, error) {
...
 			}
 			nestedProfile := p.subscription.DeepCopyObject().(*profilesv1.ProfileSubscription)
 			nestedProfile.Spec.ProfileURL = artifact.Profile.URL
 			nestedProfile.Spec.Branch = artifact.Profile.Branch
-			nestedObjs, err := New(p.ctx, nestedProfileDef, *nestedProfile, p.client, p.log).MakeOwnerlessArtifacts()
+
+			nestedSub := New(p.ctx, nestedProfileDef, *nestedProfile, p.client, p.log)
+			profileRepoName := nestedSub.profileRepo()
+			if containsKey(profileRepos, profileRepoName) {
+				return nil, fmt.Errorf("profile cannot contain profile artifact pointing to itself")
```
Callisto13, line 206: Should we re-word this error message bit? The person seeing this error is likely to not be the person who created the looping profile, so reading `"profile cannot..."` is like "well i didn't make it".

Maybe
```suggestion
				return nil, fmt.Errorf("recursive profile detected: profile cannot contain artifact pointing to itself")
```

so at least it sounds like it is not the fault of the installer? idk

bigkevmcd: Yeah, I suspect we need to output which profile is being recursively included.

In the case of _A -> B -> C -> B_ it might not be obvious what "itself" means?

aclevername: :ok: both changes sound good! I will update :smile:

pkg/profile/artifact.go
```diff
@@ -184,17 +192,21 @@ func (p *Profile) MakeOwnerlessArtifacts() ([]runtime.Object, error) {
...
 				return nil, fmt.Errorf("failed to fetch profile %q: %w", artifact.Name, err)
 			}
 			nestedProfile := p.subscription.DeepCopyObject().(*profilesv1.ProfileSubscription)
 			nestedProfile.Spec.ProfileURL = artifact.Profile.URL
 			nestedProfile.Spec.Branch = artifact.Profile.Branch
-			nestedObjs, err := New(p.ctx, nestedProfileDef, *nestedProfile, p.client, p.log).MakeOwnerlessArtifacts()
+
+			nestedSub := New(p.ctx, nestedProfileDef, *nestedProfile, p.client, p.log)
+			profileRepoName := nestedSub.profileRepo()
+			if containsKey(profileRepos, profileRepoName) {
```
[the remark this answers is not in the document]
aclevername: > Hmm, I'm on a fence of this `containsKey` if it's not being reused elsewhere.

Thats a fair point! I think I wrote it more out of anger that golang doesn't give you stuff like this for free :shakes-fist:. Although it is coming! https://github.com/golang/go/issues/45955

> Anyways, this is a preference and doesn't change the review outcome. :) I leave it up to you. :)

I'm going to keep as is (mainly out of laziness), but its a fair point raised :smile:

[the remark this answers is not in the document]
Skarlso: Soon! :D

PR #108, opened 2021-05-06
Multi-profile repo proposal

aclevername:
### Description

[rendered](https://github.com/aclevername/profiles/blob/mult-repo-proposal/docs/proposal-01-multi-profile-repo.md). Sorry in advance for the poor spelling and grammar, feel free to update :smile: 

### Checklist


aclevername:
@Callisto13 be good to get all your thoughts on this before you go on holiday :smile:

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,271 @@
...
+metadata:
+  name: bar-test
+  namespace: default
+spec:
+  profileURL: https://github.com/weaveworks/weaveworks-profiles
+  branch: main
+  profile: bar
+```
+
+This approach hides the fact that the git tag contains the profile name, and leaves it up to the profiles controller to concatinate the `profile` and `tag` value together.
```
Callisto13, line 173:
```suggestion
This approach hides the fact that the git tag contains the profile name, and leaves it up to the profiles controller to concatenate the `profile` and `tag` value together.
```

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,271 @@
...
+```
+bar/v0.10.19
+bar/v0.10.18
+foo/v0.1.1
+foo/v0.1.0
+sub/path/foo/v1.1.0
+sub/path/foo/v1.0.0
+```
+
+Here we have to different profiles, both called `foo` who are only distinguishable by there paths. If we were to write automation to parse profile repositories we might find it awkward
```
Callisto13, line 244:
```suggestion
Here we have two different profiles, both called `foo` which are only distinguishable by their paths. If we were to write automation to parse profile repositories we might find it awkward
```

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,271 @@
...
+│   └── config
+	└── main.go
+	...
+└── kyaml
+    └── main.go
+    ...
+```
+
+If you wanted to get version `v0.8.9` of `api` you would checkout to tag `api/v0.8.9` and inside the `api/` directory it would contain the desired code.
+The case of `cmd/config` is slightly less clean, its easier to write automation around detecting new profiles if the profile directory is always at the top
```
Callisto13, line 127:
```suggestion
The case of `cmd/config` is slightly less clean, it's easier to write automation around detecting new profiles if the profile directory is always at the top
```

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,271 @@
...
+kind: ProfileSubscription
+metadata:
+  name: nginx-profile-test
+  namespace: default
+spec:
+  profileURL: https://github.com/weaveworks/nginx-profile
+  branch: main
+```
+
+We want to maintain support for `branch` workflows as its useful for development, while also introducing a new `tags` approach. There are two possible approaches:
```
Callisto13, line 145:
```suggestion
We want to maintain support for `branch` workflows as it's useful for development, while also introducing a new `tags` approach. There are two possible approaches:
```

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,271 @@
...
+  name: bar-test
+  namespace: default
+spec:
+  profileURL: https://github.com/weaveworks/weaveworks-profiles
+  branch: main
+  path: bar/
+```
+
+This approach makes it clear upon inspection what tag is actually being referenced in the repository, but is less clear what directory/profile inside the repository is being referenced.
+It also maintains two completely different approach for using `branch` vs `tag`
```
Callisto13, line 202:
```suggestion
It also maintains two completely different approaches for using `branch` vs `tag`
```

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,271 @@
...
+bar/
+```
+
+```
+bar/v0.10.19
+bar/v0.10.18
+foo/v0.1.1
+foo/v0.1.0
+```
+Its clear that it has two profiles `foo` and `bar`, and that they have there respective tags. If we added support for sub-directories we introduce more complexity, for example:
```
Callisto13, line 224:
```suggestion
It's clear that it has two profiles `foo` and `bar`, and that they have their respective tags. If we added support for sub-directories we introduce more complexity, for example:
```

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,271 @@
...
+This approach makes it clear upon inspection what tag is actually being referenced in the repository, but is less clear what directory/profile inside the repository is being referenced.
+It also maintains two completely different approach for using `branch` vs `tag`
+
+## Drawbacks (Optional)
+
+- This approach to tagging repositories that contain multiple "things" is not widely used, users will likely not be familiar with it
+
+## Open Questions / Known Unknowns
+
+Support for specifying sub-directories in the tag allow more granular repository layouts, but make parsing tags less clean. For example if my repository is as follows:
```
Skarlso, line 210:
```suggestion
Support for specifying sub-directories in the tag allow more granular repository layouts, but makes parsing tags less clean. For example if my repository is as follows:
```

Skarlso:
So far so good. I would definitely opt for approach one and no sub-directories for sure.

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,271 @@
...
+
+```yaml
+apiVersion: weave.works/v1alpha1
+kind: ProfileSubscription
+metadata:
+  name: foo-test
+  namespace: default
+spec:
+  profileURL: https://github.com/weaveworks/weaveworks-profiles
+  tag: v0.1.2
```
bigkevmcd, line 157: This is really just the "version".

aclevername: yes! Will update

aclevername:
> So far so good. I would definitely opt for approach one and no sub-directories for sure.

Yes this is my preference too

docs/proposal-01-multi-profile-repo.md
```diff
@@ -0,0 +1,273 @@
...
+  name: bar-test
+  namespace: default
+spec:
+  profileURL: https://github.com/weaveworks/weaveworks-profiles
+  branch: main
+  path: bar/
+```
+
+This approach makes it clear upon inspection what tag is actually being referenced in the repository, but is less clear what directory/profile inside the repository is being referenced.
+It also maintains two completely different approaches for using `branch` vs `tag`
```
bigkevmcd, line 202: I prefer this option, path is more explicit than inferring from the tag, tagging strategies vary wildly and we can't really rely on it. Instead of a tag, you could ref a SHA (which is more reliable) and there's no path inferrable.

Skarlso: Would you mind explaining what do you mean under tagging strategies vary wildly and that we can't rely on it?
I mean we apply the strategy. We create tags and we create the folder structure for it. And if it's documented, why can't we rely on it? Do you mean that people will have difficulties adjusting to it and using it in their own repos and profiles and that we can't rely on proper adoption?

bigkevmcd: There are many tag naming policies (we are not going to be the only folks maintaining profiles).

So, it's not about _us_ relying on _our_ tag naming strategy, it's other folks... :-)

Skarlso: > Do you mean that people will have difficulties adjusting to it and using it in their own repos and profiles and that we can't rely on proper adoption?

Ok, so you're are worried about adoption part. Makes sense. But Kubernetes is widely adopted and being contributed too. We could argue that this scheme is already pretty well known?

aclevername: Adding the option for `SHA` sounds like a nice follow up feature :wink: 

> But Kustomize is widely adopted and being contributed too. We could argue that this scheme is already pretty well known?

Anecdotal but before Kevin pointed me at this repo I'd never seen this pattern before :smile: 

I think with either approach its up to us to heavily document and explain the process to the users, so it doesn't matter too much

Callisto13:
>>So far so good. I would definitely opt for approach one and no sub-directories for sure.

>Yes this is my preference too

+1 this, max depth of 1 is enough to allow for multiple profiles per repo

docs/proposal-01-multi-profile-repo.md
Callisto13: I am happy with the greater clarity of approach 2 👍

bigkevmcd:
Works for me for now :-)

aclevername:
> Works for me for now :-)

cool. Re-opening as I would like this doc to be merged so we have an historical artifact

Callisto13 (approved):

PR #113, opened 2021-05-10
Add support for version (tags) and path

aclevername:
### Description

# wip

### Checklist


controllers/profile_controller.go
```diff
@@ -95,7 +95,13 @@ func (r *ProfileSubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
-	pDef, err := git.GetProfileDefinition(pSub.Spec.ProfileURL, pSub.Spec.Branch, logger)
+	//TODO add validation for arguments in PCTL
```
Skarlso, line 98: Shouldn't we add them here too? Do you expect PCTL to ever be the only consumer / user?

Skarlso: I think we should do that here since what if someone just updates stuff by hand?

aclevername: but what will there be to update :smile: the profile subscription will be gone

PR #126, opened 2021-05-24
Add support for querying for latest version in catalog

aclevername:
### Description

You can now provide "latest" as the argument to `version`

### Checklist

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -1,9 +1,11 @@
 package catalog
 
 import (
+	"sort"
 	"strings"
 	"sync"
 
+	"github.com/hashicorp/go-version"
 	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/catalog/catalog.go
```diff
@@ -62,15 +64,47 @@ func (c *Catalog) Get(sourceName, profileName string) *profilesv1.ProfileDescrip
...
+}
+
+func getLatestVersion(profiles []profilesv1.ProfileDescription, profileName string) *profilesv1.ProfileDescription {
+	var profilesWithValidVersion []profileDescriptionWithVersion
+
+	for _, p := range profiles {
+		v, err := version.NewVersion(p.Version)
+		if err != nil {
+			continue
+		}
```
aclevername, line 97: I think we should validate a catalog entry has a valid semver version somewhere else, so here it just ignores the profile if its not valid

Callisto13: where is the somewhere else?

aclevername: ideally a validation webhook, to give the error back to the user at `kubectl create` time

Callisto13: cool 👍

Callisto13 (approved):

PR #151, opened 2021-06-07
add logging

aclevername:
### Checklist

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -2,9 +2,10 @@ package api
 
 import (
 	"encoding/json"
-	"log"
+	"fmt"
 	"net/http"
 
+	"github.com/go-logr/logr"
 	"github.com/gorilla/mux"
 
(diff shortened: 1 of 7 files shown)


pkg/api/api.go
```diff
@@ -48,62 +51,83 @@ func New(profileCatalog Catalog) API {
 // ProfilesHandler is the handler for /profiles requests.
 func (a *API) ProfilesHandler(w http.ResponseWriter, r *http.Request) {
 	query := r.URL.Query().Get("name")
+	logger := a.logger.WithValues("endpoint", "/profiles", "name", query)
```
aclevername, line 54: this makes all of the below logs pre-fixed with these key-values

bigkevmcd: we could maybe use `r.URL.Path` instead of "/profiles" ?

Skarlso (approved):
LGTM

PR #152, opened 2021-06-07
Fix api version

Skarlso:
### Checklist

### DIFF
--- a/docs/proposal-01-multi-profile-repo.md
+++ b/docs/proposal-01-multi-profile-repo.md
@@ -50,7 +50,7 @@ and as a quick reference for people coming by to learn the proposal's purpose
 without reading the entire thing.
 -->
 The purpose of this doc is to outline the proposed implementation of multi-profile repositories, and how to handle tagging profiles in this scenario.
-Currently you are limited to 1 profile per repository and you can only reference a branch in a repository. This document proposes that 
+Currently you are limited to 1 profile per repository and you can only reference a branch in a repository. This document proposes that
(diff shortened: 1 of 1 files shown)


docs/proposal-01-multi-profile-repo.md
```diff
@@ -80,7 +80,7 @@ $ tree weaveworks-profile/foo
 The `profile.yaml` would contain a reference to the artifacts relative to profile directory, for example:
 ```yaml
 $ cat foo/profile.yaml
-apiVersion: profiles.fluxcd.io/v1alpha1
+apiVersion: packages.weave.works/v1alpha1
```
aclevername, line 83: don't we need to change all the CRDs here too?

Skarlso: I was also surprised by this, but no. Group version is still weave.works.... I'm not even sure where profiles here is coming from. I think we just put that there? It could just as well be groupname/version. Regardless, this change here, does not affect code under api or examples.

Skarlso: Group is `weave.works`. And the CRD is named `  name: profilecatalogsources.weave.works` for example. That didn't reference `profiles` at all to being with. Nor fluxcd.io for that matter. So I don't think we need a code change there, but please feel free to correct me, if I see this incorrectly.

aclevername: ahhh yes your correct!

aclevername:
cut a new release after this and bump it in pctl?

Skarlso:
@aclevername Yep!

PR #158, opened 2021-06-11
remove unused func

aclevername:
### Checklist


Callisto13 (approved):

Skarlso:
I think the idea here was that this can be called outside of profiles even, by PCTL for example. Since it's on the object.

aclevername:
> I think the idea here was that this can be called outside of profiles even, by PCTL for example. Since it's on the object.

true, but the logic ended up being tested inside `pctl`. I think removing this and adding it to profiles sort of hints at the fact this probably should be defined together in 1 repo. This is the PR to move the logic to pctl https://github.com/weaveworks/pctl/pull/130/files

Issue #145, opened 2021-06-03
Support for discovering profiles in a repository


Callisto13:
Follow up from https://github.com/weaveworks/profiles/issues/127

Building a catalog currently requires manual addition of CatalogSources. This is a pain when adding a profile repo which contains multiple profiles with multiple tags.

The linked spike demonstrates how we can take a profile repo url and automatically detect new profiles and their tags and add them to the catalog.

```yaml
apiVersion: weave.works/v1alpha1
kind: ProfileCatalogSource
metadata:
  name: nginx-catalog
spec:
  repo: https://github.com/aclevername/profiles-examples
  secretRef:
     name: blueberry
```

Note: this only handles the first half of the work mentioned in the spike. Just fetch the profiles+tags of a given url, do not worry about watching repos for new tags there will be another story for this.

PR #162, opened 2021-06-14
add support for scanning repositories for profiles

aclevername:
### Description

https://github.com/weaveworks/profiles/issues/145

Merged `Lint` job into `Test` as its not worth having its own job
### Checklist


controllers/profilecatalogsource_controller.go
```diff
@@ -41,6 +48,13 @@ type ProfileCatalogSourceReconciler struct {
 // +kubebuilder:rbac:groups=weave.works,resources=profilecatalogsources/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=weave.works,resources=profilecatalogsources/finalizers,verbs=update
 
+// +kubebuilder:rbac:groups=source.toolkit.fluxcd.io,resources=gitrepositories,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=source.toolkit.fluxcd.io,resources=gitrepositories/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=source.toolkit.fluxcd.io,resources=gitrepositories/finalizers,verbs=get;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
```
aclevername, line 56: this feels horrid... but :shrug:

pkg/git/git.go
```diff
@@ -0,0 +1,53 @@
...
+
+	var tags []string
+	for _, ref := range refs {
+		if ref.Name().IsTag() {
+			tags = append(tags, ref.Name().Short())
+		}
+	}
+
+	return tags, nil
+}
```
aclevername, line 53: this doesn't have any unit tests :sob: I considered adding some, but I'd end up mocking out like 5/6 single-function interfaces. This logic gets tested as part of the acceptance tests

.github/workflows/acceptance.yaml
```diff
@@ -31,6 +34,14 @@ jobs:
           go get -u github.com/onsi/ginkgo/ginkgo
       - name: Setup Flux v2 CLI
         uses: fluxcd/flux2/action@main
+      - name: install libgit
```
Skarlso, line 37: You have to BUILD it? :O

aclevername: The version available in 'apt' isn't correct and I couldn't find a mirror :(

controllers/profilecatalogsource_controller.go
```diff
@@ -63,6 +77,26 @@ func (r *ProfileCatalogSourceReconciler) Reconcile(ctx context.Context, req ctrl
 	logger.Info("updating catalog entries", "profiles", pCatalog.Spec.Profiles)
 	r.Profiles.Update(pCatalog.Name, pCatalog.Spec.Profiles...)
 
+	gitRepoManager := gitrepository.NewManager(ctx, pCatalog.Namespace, r.Client, time.Minute*2, time.Second*5)
```
Callisto13, line 80: what are the last 2 args? can we have consts or smth for clarity?

pkg/git/git.go
Callisto13: you _could_ have a single interface with N funcs which delegate to each real function 🤷‍♀️

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/fluxcd/pkg/version"
+	sourcev1 "github.com/fluxcd/source-controller/api/v1beta1"
+	"github.com/go-logr/logr"
+	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
+	"github.com/weaveworks/profiles/pkg/gitrepository"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+//go:generate counterfeiter -o fakes/fake_git_client.go . GitClient
```
Callisto13, line 22: do you get warnings from counterfeiter when you generate with this?

they recommend now to have this line once per package:
```go
//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
```

then lines like this per interface:
```go
//counterfeiter:generate -o fakes/fake_thing.go . FooBar
```

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,171 @@
...
+	tags, err := s.gitClient.ListTags(repo.URL, secret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list tags: %w", err)
+	}
+	s.logger.Info("found tags", "url", repo.URL, "tags", tags)
+
+	var instances []gitrepository.Instance
+	for _, tag := range tags {
+		semver, path := getSemverAndPathFromTag(tag)
+		_, err := version.ParseVersion(semver)
```
Callisto13, line 74: do we care what happens if this err is not nil? am i missing the check for it?

controllers/profilecatalogsource_controller.go
aclevername: fixed

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,171 @@
...
+		if err == io.EOF {
+			return nil, nil
+		}
+
+		if err != nil {
+			return nil, fmt.Errorf("failed to read tarball file: %w", err)
+		}
+
+		switch header.Typeflag {
+		case tar.TypeReg:
```
Callisto13, line 149: could this just be an `if header.TypeFlag == tar.TypeReg`?

pkg/scanner/scanner.go
aclevername: if its not valid semver, its not a tag we care about so we ignore it

pkg/scanner/scanner.go
aclevername: I didn't notice any warnings :shrug:

Callisto13: huh 🤷‍♀️  maybe your counterfeiter is old?

aclevername: fixed! Got the warnings when I updated, good spot!

pkg/gitrepository/manager.go
```diff
@@ -0,0 +1,139 @@
...
+			return fmt.Errorf("failed to get gitrepository: %w", err)
+		}
+		if gitRes.Status.URL != "" {
+			return nil
+		}
+
+		if time.Since(startTime) > m.timeout {
+			return fmt.Errorf("timed out waiting for %s/%s gitrepository.Status.URL to be populated", gitRes.Namespace, gitRes.Name)
+		}
+		time.Sleep(m.interval)
```
Skarlso, line 85: I would much prefer using a time.After which doesn't waste as much resources as a regular sleep. That would look like this:

```go
ticker := time.NewTicker(10 * time.Second)
defer ticker.Stop()
for {
		select {
                 case <-ticker.C:
                      // do the url things
		case <-time.After(timeout):
                      // return error timeout.
                 }
}
```

Also gives much better control over timing out. And also doesn't mock around with time out after and things like that. And uses channel mechanics which are easier on the cpu than plain sleep.

pkg/gitrepository/manager.go
```diff
@@ -0,0 +1,139 @@
...
+
+func makeGitRepoName(tag, url string) string {
+	urlParts := strings.Split(url, "/")
+	repo := strings.TrimRight(urlParts[len(urlParts)-1], ".git")
+
+	splitTag := strings.Split(tag, "/")
+	if len(splitTag) == 2 {
+		return fmt.Sprintf("%s-%s-%s", repo, splitTag[0], splitTag[1])
+	}
+	return fmt.Sprintf("%s-%s", repo, tag)
```
Skarlso, line 127: We probably have to sanitise tag here, because it doesn't use domain naming conventions, right?
Like we did with GitRepository object in PCTL using sanitise. It was just doing this:

```go
var invalid = regexp.MustCompile(`[^a-zA-Z0-9_-]+`)

// SanitiseBranchName makes the branch domain name friendly.
func SanitiseBranchName(branch string) string {
	sanitized := invalid.ReplaceAllString(branch, "")
	return strings.ReplaceAll(sanitized, "_", "-")
}
```

pkg/gitrepository/manager.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+//DeleteResources deletes the gitrepository resources
+func (m *Manager) DeleteResources(gitRepos []*sourcev1.GitRepository) error {
+	for _, res := range gitRepos {
+		err := m.kClient.Delete(m.ctx, res)
+		if err != nil {
+			return err
```
Skarlso, line 135: fmt.Errorf to give context please. :) it would be nice to know which repository it couldn't delete. This goes also for the rest of the `return err`s.

pkg/gitrepository/manager.go
bigkevmcd: https://git-scm.com/docs/git-check-ref-format provides a fair bit of detail on what Git considers to be a valid branch name...and I'm pretty confident that GitHub accepts everything that Git does :-)

Skarlso: The error doesn't come from git. If we include in the kubernetes object name anything that isn't DNS compatible, like _ it will throw an error.

pkg/gitrepository/manager_test.go
```diff
@@ -0,0 +1,193 @@
...
+		})
+
+		It("creates the gitrepository resources and waits for them to be ready", func() {
+			resources, err := manager.CreateAndWaitForResources(repo, []gitrepository.Instance{
+				{
+					Tag:  "v0.1.0",
+					Path: "profile.yaml",
+				},
+				{
+					Tag:  "foo/v1.0.0",
```
Skarlso, line 66: Add invalid tags here, like, `foo_whatever` which shouldn't break things after the tag name is sanitised.

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,170 @@
...
+	gitRepositoryManager GitRepositoryManager
+	gitClient            GitClient
+	httpClient           HTTPClient
+	logger               logr.Logger
+	namespace            string
+	ctx                  context.Context
+}
+
+//New returns a Scanner
+func New(gitRepositoryManager GitRepositoryManager, gitClient GitClient, httpClient HTTPClient, logger logr.Logger) *Scanner {
```
Skarlso, line 53: Should we create a config struct? do you think this will expand?
I guess it's fine, but if we add one more we should. :D

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,170 @@
...
+	Do(req *http.Request) (*http.Response, error)
+}
+
+//Scanner for scanning repositorys
+type Scanner struct {
+	gitRepositoryManager GitRepositoryManager
+	gitClient            GitClient
+	httpClient           HTTPClient
+	logger               logr.Logger
+	namespace            string
```
Skarlso, line 48: Do I see correctly that neither namespace nor ctx is used?

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,170 @@
...
+		if err != nil {
+			s.logger.Error(err, "failed to cleanup git resources", "gitrepositories", gitRepositoryResources)
+		}
+	}()
+
+	var profiles []profilesv1.ProfileCatalogEntry
+	for _, gitRepo := range gitRepositoryResources {
+		profileDef, err := s.fetchProfileFromTarball(gitRepo)
+		if err != nil {
+			return nil, err
```
Skarlso, line 101: fmt.Errorf here too please. :)

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, fmt.Errorf("failed to GET %q: %w", gitRepo.Status.URL, err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("request failed status code %d", resp.StatusCode)
+	}
+
+	return extractProfileFromTarball(resp.Body)
+}
+
+func extractProfileFromTarball(gzipStream io.Reader) (*profilesv1.ProfileDefinition, error) {
```
Skarlso, line 132: Very cool, I like the conciseness of this function.

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,170 @@
...
+	tags, err := s.gitClient.ListTags(repo.URL, secret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list tags: %w", err)
+	}
+	s.logger.Info("found tags", "url", repo.URL, "tags", tags)
+
+	var instances []gitrepository.Instance
+	for _, tag := range tags {
+		semver, path := getSemverAndPathFromTag(tag)
+		_, err := version.ParseVersion(semver)
```
Skarlso, line 75: Super nit:
inline it into the if? 

```go
if _, err := version.ParseVersion(semver); err == nil {
...
}
```

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		if err != nil {
+			return nil, fmt.Errorf("failed to read tarball file: %w", err)
+		}
+
+		if header.Typeflag == tar.TypeReg {
+			decoder := yaml.NewYAMLOrJSONDecoder(tarReader, 10000)
+			var profileDef profilesv1.ProfileDefinition
+			err = decoder.Decode(&profileDef)
```
Skarlso, line 152: Super nit:
Inline into the below if?

pkg/scanner/scanner.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	gitRepositoryResources, err := s.gitRepositoryManager.CreateAndWaitForResources(repo, instances)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create gitrepository resources: %w", err)
+	}
+	s.logger.Info("gitrepositorys created", "gitrepositories", gitRepositoryResources)
+
+	defer func() {
+		err := s.gitRepositoryManager.DeleteResources(gitRepositoryResources)
```
Skarlso, line 91: Super nit:
Inline into the below if? :D

pkg/gitrepository/manager.go
```diff
@@ -0,0 +1,139 @@
...
+
+	splitTag := strings.Split(tag, "/")
+	if len(splitTag) == 2 {
+		return fmt.Sprintf("%s-%s-%s", repo, splitTag[0], splitTag[1])
+	}
+	return fmt.Sprintf("%s-%s", repo, tag)
+}
+
+//DeleteResources deletes the gitrepository resources
+func (m *Manager) DeleteResources(gitRepos []*sourcev1.GitRepository) error {
```
Skarlso, line 131: This is missing tests. :)

pkg/gitrepository/manager.go
```diff
@@ -0,0 +1,139 @@
...
+	Create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error
+	Delete(ctx context.Context, obj client.Object, opts ...client.DeleteOption) error
+}
+
+//NewManager returns a Manager struct
+func NewManager(ctx context.Context, namespace string, kClient Kubernetes, timeout, interval time.Duration) *Manager {
+	return &Manager{
+		kClient:   kClient,
+		namespace: namespace,
+		ctx:       context.TODO(),
```
Skarlso, line 44: you aren't using the passed in context. This should be `ctx`.

pkg/gitrepository/manager.go
```diff
@@ -0,0 +1,139 @@
...
+
+//Instance contains a tag and path of profile.yaml
+type Instance struct {
+	Tag  string
+	Path string
+}
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+//counterfeiter:generate -o fakes/fake_kubernetes.go . Kubernetes
+//Kubernetes interface for itneracting with kubernetes
```
Skarlso, line 32:
```suggestion
//Kubernetes interface for interacting with Kubernetes
```

api/v1alpha1/profilecatalogsource_types.go
```diff
@@ -19,13 +19,30 @@ package v1alpha1
...
+	// The secret name containing the Git credentials.
+	// For HTTPS repositories the secret must contain username and password
+	// fields.
+	// For SSH repositories the secret must contain identity, identity.pub and
+	// known_hosts fields.
+	// +optional
+	SecretRef *meta.LocalObjectReference `json:"secretRef,omitempty"`
 }
 
 // ProfileDescription defines details about a given profile.
```
Skarlso, line 48: err, I know it's not your concern, but I just noticed that this isn't the right description for this struct. :D Would you mind? Pretty please? 🙇

controllers/profilecatalogsource_controller.go
```diff
@@ -63,6 +77,28 @@ func (r *ProfileCatalogSourceReconciler) Reconcile(ctx context.Context, req ctrl
...
+			secret = &corev1.Secret{}
+			err := r.Client.Get(ctx, client.ObjectKey{Name: repo.SecretRef.Name, Namespace: req.Namespace}, secret)
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to find secret for repo %v: %w", repo, err)
+			}
+		}
+
+		profiles, err := scanner.ScanRepository(repo, secret)
+		if err != nil {
+			return ctrl.Result{}, err
```
Skarlso, line 97: Please add some logging here, so we know that's happening.

controllers/profilecatalogsource_controller.go
```diff
@@ -63,6 +77,28 @@ func (r *ProfileCatalogSourceReconciler) Reconcile(ctx context.Context, req ctrl
...
+	timeout := time.Minute * 2
+	interval := time.Second * 5
+	gitRepoManager := gitrepository.NewManager(ctx, pCatalog.Namespace, r.Client, timeout, interval)
+	scanner := scanner.New(gitRepoManager, &git.Client{}, http.DefaultClient, logger)
+	for _, repo := range pCatalog.Spec.Repos {
+		logger.Info("scan repo for profiles", "repo", repo)
+		var secret *corev1.Secret
+		if repo.SecretRef != nil {
+			secret = &corev1.Secret{}
+			err := r.Client.Get(ctx, client.ObjectKey{Name: repo.SecretRef.Name, Namespace: req.Namespace}, secret)
```
Skarlso, line 89: Super nit: Inline this please? :D

pkg/git/git.go
```diff
@@ -0,0 +1,53 @@
...
+	var auth transport.AuthMethod
+	if secret != nil {
+		authStrategy, err := gogit.AuthSecretStrategyForURL(url)
+
+		if err != nil {
+			return nil, err
+		}
+		authMethod, err := authStrategy.Method(*secret)
+		if err != nil {
+			return nil, err
```
Skarlso, line 32: fmt.Errorf

pkg/git/git.go
```diff
@@ -0,0 +1,53 @@
+package git
```
Skarlso, line 1: This package doesn't appear to have any tests? I assume you are writing them? :D

Skarlso: Oh I see the resolved conversation. Sorry.

pkg/git/git.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"github.com/fluxcd/source-controller/pkg/git/gogit"
+	extgogit "github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+
```
Skarlso, line 7: remove extra linebreak

api/v1alpha1/profilecatalogsource_types.go
```diff
@@ -19,13 +19,30 @@ package v1alpha1
...
 type ProfileCatalogSourceSpec struct {
 	// Profiles is the list of profiles exposed by the catalog
+	// +optional
 	Profiles []ProfileCatalogEntry `json:"profiles,omitempty"`
+	// Repositories is the list of repositories to scan for profiles
+	// +optional
+	Repos []Repository `json:"repositories,omitempty"`
+}
+
+type Repository struct {
```
Skarlso, line 36: Missing description.

api/v1alpha1/profilecatalogsource_types.go
```diff
@@ -19,13 +19,30 @@ package v1alpha1
...
+	// URL is the URL of the repository
```
Skarlso, line 37: Can we expand on this? Like, what format, can it be private, public, ssh or https...

api/v1alpha1/profilecatalogsource_types.go
```diff
@@ -19,13 +19,30 @@ package v1alpha1
...
+	"github.com/fluxcd/pkg/apis/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // ProfileCatalogSourceSpec defines the desired state of ProfileCatalogSource
 type ProfileCatalogSourceSpec struct {
 	// Profiles is the list of profiles exposed by the catalog
+	// +optional
 	Profiles []ProfileCatalogEntry `json:"profiles,omitempty"`
+	// Repositories is the list of repositories to scan for profiles
```
Skarlso, line 31:
```suggestion
	// Repos contains a list of repositories to scan for profiles
```

api/v1alpha1/profilecatalogsource_types.go
```diff
@@ -19,13 +19,30 @@ package v1alpha1
...
+	// +optional
+	Repos []Repository `json:"repositories,omitempty"`
+}
+
+type Repository struct {
+	// URL is the URL of the repository
+	URL string `json:"url,omitempty"`
+	// The secret name containing the Git credentials.
+	// For HTTPS repositories the secret must contain username and password
```
Skarlso, line 40: What about tokens? Should we include that or just say, in case of tokens the password is the token and the username is not required?

Skarlso (changes requested):
small things really only... 

Well done! This is an awesome feature :)

pkg/gitrepository/manager.go
aclevername: I think since this is using the tag format (string/semver) not an arbitrary branch name it should be okay?

pkg/gitrepository/manager.go
aclevername: oops good spot!

pkg/gitrepository/manager.go
aclevername: doesn't this mean we would hammer the k8s API? I introduced the delay on purpose, to prevent throttling.

Happy to change it to use a select statement all the same, but having a sleep after the k8s api call seems unavoidable?

pkg/gitrepository/manager.go
aclevername: :joy:

pkg/scanner/scanner.go
aclevername: I think its okay. `httpClient` and `gitCleint` are only exposed for testing. I could switch to the `export_test` method instead to reduce this to 2 arguments

api/v1alpha1/profilecatalogsource_types.go
aclevername: is it not? I copied from flux https://github.com/fluxcd/source-controller/blob/2dc713258d3c4cae91923dc9689e6b468de1d54e/api/v1beta1/gitrepository_types.go#L42-L48

controllers/profilecatalogsource_controller.go
aclevername: returning an err does produce a log, and we log above whats happening, so I think its okay?

api/v1alpha1/profilecatalogsource_types.go
aclevername: huh linter didn't complain, weird

pkg/scanner/scanner.go
aclevername: this is an error from a local func, so its up to the local func to correctly format the error message (this is a pattern I tend to stick to). If it was calling an external package I would wrap it :smile:

pkg/scanner/scanner.go
aclevername: left over from when this func did k8s calls! good spot :smile:

api/v1alpha1/profilecatalogsource_types.go
aclevername: don't think a token is supported https://fluxcd.io/docs/components/source/gitrepositories/#https-authentication

aclevername: In our docs we will just link to the flux docs, since we are mirroring them in authentication

api/v1alpha1/profilecatalogsource_types.go
Skarlso: ? This is comment is for the struct below this comment. :D

aclevername: oooh lol

pkg/gitrepository/manager.go
Skarlso: No, it doesn't. The select blocks. It's not going to call the k8s api continuously. But the ticker sucks I decided. The context with timeout is a LOT nicer. So use this instead which is much more readable if you like:

```go
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	for {
		select {
		case <-ctx.Done():
			fmt.Println("timeout")
			return err
		case <-time.After(interval):
			// do the thing in this interval
		}
	}
```

Example code: https://play.golang.org/p/SH8_uW9cUHy

pkg/gitrepository/manager.go
Skarlso: Yeah, sure. If you think it couldn't happen, I'm happy to ignore. :)

pkg/scanner/scanner.go
Skarlso: If the calling function properly handles and tells you what the problem was, then it's fine. Otherwise, you wouldn't know what failed and which git resource, right?

controllers/profilecatalogsource_controller.go
Skarlso: oh okay. :)

api/v1alpha1/profilecatalogsource_types.go
Skarlso: Ah okay

pkg/gitrepository/manager.go
aclevername: like it!

pkg/scanner/scanner.go
aclevername: Yeah! Here it passes in all the info it needs for it to correctly error. 

is only in a func for making the code cleaner, it could/should be able to get copied and pasted in line, and none of the error messages change :smile:

Skarlso: Cool 👍

pkg/gitrepository/manager.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (m *Manager) waitForURL(gitRes *sourcev1.GitRepository) error {
+	ctx, cancel := context.WithTimeout(context.Background(), m.timeout)
+	defer cancel()
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("timed out waiting for %s/%s gitrepository.Status.URL to be populated", gitRes.Namespace, gitRes.Name)
+		case <-time.After(m.interval):
+			err := m.kClient.Get(m.ctx, client.ObjectKeyFromObject(gitRes), gitRes)
```
Skarlso, line 79: You could even go crazy and ignore the this case in here, and put it in the beginning of the for loop then leave this empty. Because it's literally only there to progress the for loop. So like this:

```go
	for {
		select {
		case <-ctx.Done():
			return fmt.Errorf("timed out waiting for %s/%s gitrepository.Status.URL to be populated", gitRes.Namespace, gitRes.Name)
		case <-time.After(m.interval):
		}      
		if err := m.kClient.Get(m.ctx, client.ObjectKeyFromObject(gitRes), gitRes); err != nil {
			return fmt.Errorf("failed to get gitrepository: %w", err)
		}
		if gitRes.Status.URL != "" {
			return nil
		}
	}
```

aclevername: I prefer it as is :smile:

Skarlso: Okay. :D

Skarlso (approved):
LGTM! Well done! :)

PR #163, opened 2021-06-15
Add DefaultValues field for authors

Callisto13:
### Description

First part of https://github.com/weaveworks/pctl/pull/143.

Totally open for naming changes.
I went for a simple `string` rather than a `map[string]interface{}` (controller-gen hates that since it is basically impossible to `DeepCopy`), or `apiextensionsv1.JSON` since I hate that thing. This means the author can define values ConfigMap style (it will get passed directly into a ConfigMap anyway):

```yaml
...
  artifacts:
    - name: nginx-server
      chart:
        path: nginx/chart
        defaultValues: |
          replicaCount: 3
```

See comment [here](https://github.com/weaveworks/pctl/issues/58#issuecomment-850494433) on why I wasn't into having them write a file for each valueset at the time I started this. The game may have changed since then, but I feel there is also a wider conversation coming about "values" in general so idk.

Also removed the `Values` field, since nobody can even use it 🤷‍♀️ .

### Checklist


Skarlso (approved):
LGTM

config/crd/bases/weave.works_profiledefinitions.yaml
```diff
@@ -46,6 +46,11 @@ spec:
                       description: Chart defines properties to access a remote chart.
                         This is an optional value. It is ignored in case Path is defined.
                       properties:
+                        defaultValues:
+                          description: DefaultValues holds the default values for
+                            this Helm release Artifact. These can be overridden by
+                            the user, but will otherwise apply.
+                          type: string
```
aclevername, line 53: string? I was expecting a map[string]interface or something. What does it look like to configure this field?

Callisto13: did you even read the description tho?

aclevername: NO THATS FOR LOSERS

aclevername (approved):

api/v1alpha1/profile_types.go
```diff
@@ -89,6 +89,10 @@ type Chart struct {
 	// This is an optional value. If defined, it takes precedence over other Chart fields.
 	// +optional
 	Path string `json:"path,omitempty"`
+	// DefaultValues holds the default values for this Helm release Artifact.
+	// These can be overridden by the user, but will otherwise apply.
+	// +optional
+	DefaultValues *string `json:"defaultValues,omitempty"`
```
aclevername, line 95: why a pointer?

Callisto13: that... is an excellent question 😂

PR #166, opened 2021-06-17
Bump deps

Callisto13:


.github/workflows/test.yaml
```diff
@@ -12,6 +12,9 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.16'
```
aclevername, line 17: add it to acceptance too?

aclevername (approved):

.github/workflows/test.yaml
Callisto13: they seem fine? but sure, why not 🤷‍♀️

PR #184, opened 2021-06-24
Switch to grpc-gateway on the manager and add graceful shutdown support for all services

### LINKED ISSUE #165: Switch to gRPC gateway API
**What feature/behavior/change do you want?**
Currently, we have a hand-rolled path-based API, we are standardising on gRPC via grpc-gateway.

We should define an API proto and move our API code to use it.

**Why do you want this feature?**

API standardisation!

Essentially we want to have standards around the APIs exposed by our components to make it easier to build clients.

Skarlso:
### Description

Closes #165 

### Checklist
- [x] Added/modified documentation as required (such as the `README.md` (diagrams, usage, roadmap etc), and anything in `examples/`)


pkg/catalog/catalog.go
```diff
@@ -112,14 +112,19 @@ func (c *Catalog) ProfilesGreaterThanVersion(logger logr.Logger, sourceName, pro
 	if !ok {
 		return nil
 	}
+	logger.Info("Parsing profilesVersion", "version", profileVersion)
 	cv, err := version.ParseVersion(profileVersion)
 	if err != nil && profileVersion != "latest" {
 		return nil
 	}
+	var count int
```
Skarlso, line 120: ops, this is debugging

api/v1alpha1/profilecatalogsource_types.go
```diff
@@ -58,7 +58,7 @@ type ProfileCatalogEntry struct {
 	Tag string `json:"tag,omitempty"`
 	// CatalogSource is the name of the catalog the profile is listed in
 	// +optional
-	CatalogSource string `json:"catalog,omitempty"`
+	CatalogSource string `json:"catalogSource,omitempty"`
```
aclevername, line 61: breaking changes :ahhhhh:

main.go
```diff
@@ -112,17 +128,105 @@ func main() {
...
+		}
+		return nil
+	})
+
+	if err := g.Wait(); err != nil {
+		if errors.Is(err, context.Canceled) {
+			setupLog.Error(err, "context was cancelled")
+		} else {
+			setupLog.Error(err, "server error detected")
+		}
```
aclevername, line 229: so much "magic" :cry:

pkg/api/api.go
```diff
@@ -23,118 +24,108 @@ type Catalog interface {
...
+	sourceName := request.GetSourceName()
+	profileName := request.GetProfileName()
+	logger := p.logger.WithValues("func", "Get", "catalog", sourceName, "profile", profileName)
 	if sourceName == "" || profileName == "" {
-		a.logger.Error(fmt.Errorf("missing query param"), "profile and/or catalog not set")
-		a.logAndWriteHeader(w, http.StatusBadRequest)
-		return
+		errMsg := fmt.Errorf("missing query param: sourceName: %q, profileName: %q", sourceName, profileName)
+		logger.Error(errMsg, "profile and/or catalog not set")
+		return nil, status.Errorf(codes.InvalidArgument, errMsg.Error())
```
bigkevmcd, line 50: 💯

pkg/protos/transform.go
```diff
@@ -0,0 +1,25 @@
+package protos
+
+import profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
+
+// TransformCatalogEntry takes a Kubernetes catalog entry and creates a proto catalog entry out of it.
```
aclevername, line 5:
```suggestion
// TransformCatalogEntry takes a profilesv1 catalog entry and creates a proto catalog entry out of it.
```

pkg/protos/transform.go
```diff
@@ -0,0 +1,25 @@
...
+		CatalogSource: origin.CatalogSource,
+		Url:           origin.URL,
+		Name:          origin.ProfileDescription.Name,
+		Description:   origin.ProfileDescription.Description,
+		Maintainer:    origin.ProfileDescription.Maintainer,
+		Prerequisites: origin.ProfileDescription.Prerequisites,
+	}
+}
+
+// TransformCatalogEntryList takes a slice of Kubernetes catalog entries and creates a proto catalog entry slice out of it.
```
aclevername, line 18:
```suggestion
// TransformCatalogEntryList takes a slice of profilesv1 catalog entries and creates a proto catalog entry slice out of it.
```

main.go
Skarlso: Yeah I wish it could be nicer, but this is almost boilerplate graceful shutdown handling core with proper timeouts so it doesn't hang and it gets impossible to kill. I'm tinkering with extracting it, but that kills any defer cancels that we might call here.

api/v1alpha1/profilecatalogsource_types.go
Skarlso: The breaking change was to not update this when the variable was renamed hahahahaha

main.go
Skarlso: Actually, my tinkering turned out some pretty awesome stuff! :D 
So stayed tuned for tomorrow, I'll show you the new code. :D Hopefully you'll like that more. :))))))

README.md
```diff
@@ -93,6 +95,21 @@ explained in `pctl`'s README section [Working with Profiles](https://github.com/
 
 The dev tag's format is as follows: `<currentLatestReleaseTag>-<branch-name>`.
 
+### Generating the protocol files
+
+`profiles` is using [buf](https://buf.build/) to generate and lint protobuf implementations. The protocol descriptor file
+can be found here [profiles.proto](./proto/profiles.proto). To update these files, install `buf` and run the make target
+`make generate-protoc`. Profiles, using the [tools](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) way to track dependencies
```
Callisto13, line 102:
```suggestion
`make generate-protoc`. Profiles, using the [tools](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) way to track dependencies,
```

api/v1alpha1/profilecatalogsource_types.go
Callisto13: 🙉

pkg/manager/server.go
```diff
@@ -0,0 +1,44 @@
...
+		logger:  logger,
+		manager: manager,
+	}
+}
+
+// Start starts the manager server.
+func (s *Server) Start(ctx context.Context) error {
+	shutdownContext, cancel := context.WithCancel(context.Background())
+	s.cancel = cancel
+	g, _ := errgroup.WithContext(ctx)
```
bigkevmcd, line 30: Should this use the returned context?

Makefile
```diff
@@ -157,8 +157,22 @@ docker-push: ## Push the docker image
...
+	GO111MODULE=off go get sigs.k8s.io/mdtoc || true
+
 ##@ Utilities
 
+generate-protoc: ## Generate the protobuf files
+	buf generate
+
+lint-protoc: ## lint the protocol files
+	buf lint
+
```
aclevername, line 175: do these need to be called as part of `make lint` and `make generate`?

Skarlso: Hm, lint, maybe. The generate is not needed. Only if you change profiles.proto.

README.md
```diff
@@ -93,6 +95,21 @@ explained in `pctl`'s README section [Working with Profiles](https://github.com/
...
 
+### Generating the protocol files
+
+`profiles` is using [buf](https://buf.build/) to generate and lint protobuf implementations. The protocol descriptor file
+can be found here [profiles.proto](./proto/profiles.proto). To update these files, install `buf` and run the make target
+`make generate-protoc`. Profiles, using the [tools](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) way to track dependencies,
+and should already have all the necessary protoc tools installed.
+
+When running `buf` for the first time, install all the necessary libraries by running `buf beta mod update`. This will install
+dependencies defined in [buf.yaml](./buf.yaml) file.
```
aclevername, line 106: we have a `dependencies` part of the README, can you add buf to it?

api/v1alpha1/profilecatalogsource_types.go
aclevername: :upside_down_face:

pkg/api/api.go
```diff
@@ -23,118 +24,106 @@ type Catalog interface {
...
-	a.logger.Info(fmt.Sprintf("returning %d", statusCode), "statuscode", statusCode)
-	w.WriteHeader(statusCode)
+var _ protos.ProfilesServiceServer = &ProfilesCatalogService{}
+
+// NewCatalogAPI returns a profiles catalog api implementation.
+func NewCatalogAPI(profileCatalog Catalog, logger logr.Logger) *ProfilesCatalogService {
+	return &ProfilesCatalogService{
+		profileCatalog: profileCatalog,
+		logger:         logger,
+	}
```
aclevername, line 128: nit: can we have the `New` at the top?

pkg/manager/server.go
Skarlso: Nope, because the shutdown context needs to be separate from that. Yet, the passed in context is still important, because if anything else using that context from the manager dies, the context needs to be able to signal everything else that it died. So if one of the servers that the handler handles dies, everyone is notified and restarted. The assumption is that these are co-dependent servers. Now, if that is not the case, then just don't use the passed in context. Right now, the returned context here, has no use, as the server has its own shutdown context which it needs to keep track of separately to call with `Stop` to be conformant with the Service interface.

aclevername (approved):
LGTM, couple comments. Good sign that the acceptance tests barely had to change :smile:

PR #185, opened 2021-06-24
only re-scan new tags

aclevername:
### Description

This PR prevents every reconciliation loop from causing all tags to be re-scanned. Instead it now stores a history of the tags its scanned in its status, and does not re-scan them if they change.

We also have to think about the fact that the cache is in-memory, so we could have a scenario where:
- The status contains a list of tags scanned
- The cache is empty, as the container got restarted

I've added a `CatalogExists` check to check for this scenario, when this happens it ignores the status and does a full re-scan.

### Checklist


controllers/profilecatalogsource_controller.go
```diff
@@ -91,17 +95,43 @@ func (r *ProfileCatalogSourceReconciler) Reconcile(ctx context.Context, req ctrl
...
 			return ctrl.Result{}, err
 		}
-		r.Profiles.Update(pCatalog.Name, profiles...)
+
+		updateScannedRepositoryStatus(&pCatalog, repo, newTags)
+		logger.Info("updating catalog with scanning reuslts", "profiles", profiles)
+		r.Profiles.Append(pCatalog.Name, profiles...)
+	}
+
+	if err := r.Client.Status().Update(ctx, &pCatalog); err != nil {
```
Skarlso, line 115: Do a Get before updating, otherwise this might fail with a version collision.

pkg/catalog/catalog.go
```diff
@@ -25,8 +25,19 @@ func New() *Catalog {
 	}
 }
 
-// Update updates the catalog by replacing existing profiles with new profiles
-func (c *Catalog) Update(sourceName string, profiles ...profilesv1.ProfileCatalogEntry) {
+// Append appends the existing profiles with new profiles
```
Skarlso, line 28: -appends. The verb is already Append. :) I think that reads nicely.
// Append the existing profiles to the new profiles.

Skarlso (commented):
LGTM, one question and one remark

api/v1alpha1/profilecatalogsource_types.go
```diff
@@ -67,6 +67,16 @@ type ProfileCatalogEntry struct {
...
+	ScannedRepositories []ScannedRepository `json:"scannedRepositories,omitempty"`
+}
+
+// ScannedRepository contains the list of repositories that have been scanned and
+// what tags have bene processed
+type ScannedRepository struct {
+	// URL is the repository URL
+	URL string `json:"url,omitempty"`
+	// Tags is the list of tags that have been scanned
+	Tags []string `json:"tags,omitempty"`
```
aclevername, line 79: I feel like this could have a better name, LMK if anyone has one :smile:

bigkevmcd: Looks good enough to me, reflects what it is :-)

Skarlso:
This creates a GitRepository for every repository, right? Even those that come with a catalog entry? Is that okay?

aclevername:
> This creates a GitRepository for every repository, right? Even those that come with a catalog entry? Is that okay?

I don't follow :confused: could you expand up on what you mean?

Skarlso:
> > This creates a GitRepository for every repository, right? Even those that come with a catalog entry? Is that okay?
> 
> I don't follow 😕 could you expand up on what you mean?

Never mind. :D I misunderstood something. :D

controllers/profilecatalogsource_controller.go
```diff
@@ -74,32 +90,91 @@ func (r *ProfileCatalogSourceReconciler) Reconcile(ctx context.Context, req ctrl
...
+		}
+
+		profiles, newTags, err := scanner.ScanRepository(repo, secret, alreadyScannedTags)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
-		r.Profiles.Update(pCatalog.Name, profiles...)
+
+		if catalogExists {
+			createOrAppendScannedRepositoryStatus(&pCatalog, repo, newTags)
```
Skarlso, line 130: These two functions are doing the same thing. I think it would be a lot nicer to have a single function with a parameter called `update`. Which if true, then just append. Otherwise, let it do its thing. So like:

```go
func updateScannedRepositoryStatus(pCatalog *profilesv1.ProfileCatalogSource, repo profilesv1.Repository, newTags []string, replace bool) {
	for i, scannedRepo := range pCatalog.Status.ScannedRepositories {
		if scannedRepo.URL == repo.URL {
                         if replace {
			      pCatalog.Status.ScannedRepositories[i].Tags = newTags
                         } else {
                               pCatalog.Status.ScannedRepositories[i].Tags = append(scannedRepo.Tags, newTags...)
                         }
			return
		}
	}
	pCatalog.Status.ScannedRepositories = append(pCatalog.Status.ScannedRepositories, profilesv1.ScannedRepository{
		URL:  repo.URL,
		Tags: newTags,
	})
}
```

So you can remove the other function. WDYT?

Skarlso (approved):
Approved with a single remark. :)

controllers/profilecatalogsource_controller.go
aclevername: done!

PR #186, opened 2021-06-25
Remove the logger from the interface methods

Skarlso:
### Description

https://github.com/weaveworks/pctl/issues/166

### Checklist
- [x] Added/modified documentation as required (such as the `README.md` (diagrams, usage, roadmap etc), and anything in `examples/`)

### DIFF
--- a/.github/workflows/pr-labels.yaml
+++ b/.github/workflows/pr-labels.yaml
@@ -13,7 +13,7 @@ jobs:
         env:
           LABELS: ${{ toJson(github.event.pull_request.labels.*.name) }}
         run: |
-          if ! jq --exit-status 'any(test("kind/feature") or test("kind/docs") or test("kind/bug") or test("skip-release-notes"))' >/dev/null <<< $LABELS; then
-            echo "::error ::Please set either kind/feature, kind/bug or skip-release-notes as label"
+          if ! jq --exit-status 'any(test("kind/feature") or test("kind/improvement") or test("kind/docs") or test("kind/bug") or test("skip-release-notes"))' >/dev/null <<< $LABELS; then
+            echo "::error ::Please set either kind/feature, kind/bug, kind/improvement, kind/docs or skip-release-notes as label"
             exit 1
           fi
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -87,7 +87,7 @@ var _ = BeforeSuite(func() {
 	})
 	Expect(err).ToNot(HaveOccurred())
 
-	profiles := catalog.New()
+	profiles := catalog.New(logf.Log)
 	catalogReconciler = &controllers.ProfileCatalogSourceReconciler{
 		Client:   k8sManager.GetClient(),
 		Log:      ctrl.Log.WithName("controllers").WithName("profilecatalog"),
--- a/main.go
+++ b/main.go
@@ -25,10 +25,11 @@ import (
 	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
 	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
 	sourcev1 "github.com/fluxcd/source-controller/api/v1beta1"
-	"github.com/weaveworks/profiles/pkg/api"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+
+	"github.com/weaveworks/profiles/pkg/api"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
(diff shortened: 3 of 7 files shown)


aclevername (changes requested):
Its passed in rather than set on the catalog because the logger gets modified for each catalog source https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/controllers/profilecatalogsource_controller.go#L63

bigkevmcd:
> Its passed in rather than set on the catalog because the logger gets modified for each catalog source
> 
> https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/controllers/profilecatalogsource_controller.go#L63

That's not quite what's happening there, when you call:

```go
logger := r.Log.WithValues("profilecatalogsource", req.NamespacedName) 
```

You get a clone of the original logger, but with additional log context values.

It's perfectly safe to do this.

Skarlso:
> Its passed in rather than set on the catalog because the logger gets modified for each request
> 
> https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/pkg/api/api.go#L94

I see, good point. But this can be shifted into the function itself. Which also makes more sense. The function itself can give more context.

Skarlso:
> > Its passed in rather than set on the catalog because the logger gets modified for each request
> > https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/pkg/api/api.go#L94
> 
> I see, good point. But this can be shifted into the function itself. Which also makes more sense. The function itself can give more context.

Also, to be fair, this whole thing changed in #184 a bit.

aclevername:
> > Its passed in rather than set on the catalog because the logger gets modified for each catalog source
> > https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/controllers/profilecatalogsource_controller.go#L63
> 
> That's not quite what's happening there, when you call:
> 
> ```go
> logger := r.Log.WithValues("profilecatalogsource", req.NamespacedName) 
> ```
> 
> You get a clone of the original logger, but with additional log context values.
> 
> It's perfectly safe to do this.
> 
> EDIT: https://github.com/go-logr/logr/blob/master/logr.go#L236

I updated my comment as the link was wrong. I'm not sure what your point your trying to convey :smile: ? I wasn't saying it was dangerous, just that its intentional that we pass the logger in. Its useful as it contains more context (more key:values) about the request. If we set the logger on the catalog struct and don't pass it in each time we don't gain this new context.

aclevername:
> > Its passed in rather than set on the catalog because the logger gets modified for each request
> > https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/pkg/api/api.go#L94
> 
> I see, good point. But this can be shifted into the function itself. Which also makes more sense. The function itself can give more context.

I am not an expert structured logger (or an expert at anything :laughing: ), but I think it makes more sense to set it at the request entry point? It sort of makes the whole flow quite clean as all the context is set at the very beginning. But Its not a hill I'm going to die on, so feel free to refactor :smile:

Skarlso:
> > > Its passed in rather than set on the catalog because the logger gets modified for each request
> > > https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/pkg/api/api.go#L94
> > 
> > 
> > I see, good point. But this can be shifted into the function itself. Which also makes more sense. The function itself can give more context.
> 
> I am not an expert structured logger (or an expert at anything 😆 ), but I think it makes more sense to set it at the request entry point? It sort of makes the whole flow quite clean as all the context is set at the very beginning. But Its not a hill I'm going to die on, so feel free to refactor 😄

I get your point, I do. But then the others should also have logging? :D Like, it's weird that only two out of the five have it. :D

aclevername:
> > > > Its passed in rather than set on the catalog because the logger gets modified for each request
> > > > https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/pkg/api/api.go#L94
> > > 
> > > 
> > > I see, good point. But this can be shifted into the function itself. Which also makes more sense. The function itself can give more context.
> > 
> > 
> > I am not an expert structured logger (or an expert at anything laughing ), but I think it makes more sense to set it at the request entry point? It sort of makes the whole flow quite clean as all the context is set at the very beginning. But Its not a hill I'm going to die on, so feel free to refactor smile
> 
> I get your point, I do. But then the others should also have logging? :D Like, it's weird that only two out of the five have it. :D

I think thats purely because they don't have anything "worth" logging. Only the version comparisions have errors that can occur, hence the logger being passed in :smile:

Skarlso:
> > > > > Its passed in rather than set on the catalog because the logger gets modified for each request
> > > > > https://github.com/weaveworks/profiles/blob/ef740a45e3038e476de957fa2bc5b3be14c7676f/pkg/api/api.go#L94
> > > > 
> > > > 
> > > > I see, good point. But this can be shifted into the function itself. Which also makes more sense. The function itself can give more context.
> > > 
> > > 
> > > I am not an expert structured logger (or an expert at anything laughing ), but I think it makes more sense to set it at the request entry point? It sort of makes the whole flow quite clean as all the context is set at the very beginning. But Its not a hill I'm going to die on, so feel free to refactor smile
> > 
> > 
> > I get your point, I do. But then the others should also have logging? :D Like, it's weird that only two out of the five have it. :D
> 
> I think thats purely because they don't have anything "worth" logging. Only the version comparisions have errors that can occur, hence the logger being passed in 😄

Fiinnnneee. But I'm fixing the frigging pr label checker :D

PR #201, opened 2021-07-02
Bump all the deps

Callisto13:


aclevername (approved):
shouldn't you be writing docs :trollface:

Issue #144, opened 2021-06-02
Profile Developer guide documentation


aclevername:
Lets document how to write a profile, providing examples covering all of the different features available.
- Kustomize artifacts
- Helm artifacts
  - Helm values provided during installation time
  - Helm values defined in the profile definition
  - Local charts
  - Remote charts
- Nested profile artifacts
- Versioning (and how the catalog detects versions)

Callisto13:
We have decided that for now (and perhaps for always 🤷‍♀️ ) we will put our docs on the new `profiles.dev` domain.

https://docusaurus.io/ is the hot new thing everyone is using, so let's go with that too to be consistent.

When writing the docs, let's try to very very clearly delineate the 3 different users we have:
- Profile author
- Catalog manager
- Profile user

Callisto13:
- [x] intro and basics tutorial (https://github.com/weaveworks/profiles/pull/195)
  - the basics tutorial will deliberately leave out a lot of stuff and only talk about the bare minimum users need to begin
- [ ] highly detailed author docs
  - will include versioning etc
- [ ] highly detailed installer docs
  - will include pctl and catalog stuff
- [ ] catalog manager docs
- [ ] autogen pctl commands
- [ ] run all example snippets
- [ ] troubleshooting
- [ ] faqs
- [ ] advanced? 🤷‍♀️ 
- [ ] change logos and pics from the default
- [ ] replace all schemas with autogenerated one
- [ ] deploy somewhere

Skarlso:
Please put any documentation things for profiles into https://github.com/weaveworks/profiles-userdocs/ from now on.

PR #228, opened 2021-07-08
Autogenerate pctl commands in docs

Callisto13:
### Description

- removes the hardcoded helps
- generates them automatically on serve
- they are gitignored and will no longer be committed
- uses new pctl docgen command

~Blocked until corresponding PR to pctl is merged and a new release is out: https://github.com/weaveworks/pctl/pull/207~

Unblocked

I added this note to the general docs pr https://github.com/weaveworks/profiles/issues/144#issuecomment-871262933:
> move the docs out of the profiles repo? or smth. they should be deployed whenever there are new releases in either pctl or profiles, so whatever makes it easier and more intuitive to do that idk


.github/workflows/build-docs.yaml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Fetch latest pctl binary
+        run: |
+          curl -sL https://api.github.com/repos/weaveworks/pctl/releases/latest 2>/dev/null | \
+          jq -r '.assets[] | select(.browser_download_url | contains("linux_amd64")) | .browser_download_url' | \
+          xargs curl -sL | sudo tar xvz -C /usr/local/bin && chmod +x /usr/local/bin/pctl && pctl --help
```
Skarlso, line 19: While it makes sense, this now adde another connection to pctl from profiles. It was pctl->profiles now it's pctl<->profiles. Guess, what I'm trying to say is, that we are getting closer and closer to just pull pctl into profiles. :)

.github/workflows/deploy-docs.yaml
```diff
@@ -16,6 +16,11 @@ jobs:
       - uses: webfactory/ssh-agent@v0.5.0
         with:
           ssh-private-key: ${{ secrets.PROFILES_DOCS_WEAVEWORKS_DOCS_BOT_DEPLOY_KEY }}
+      - name: Fetch latest pctl binary
```
Skarlso, line 19: While I understand why these are separate, this adds two downloads. Does it make sense to keep these as completely separate actions, not just like one action with two steps?

Skarlso (approved):
LGTM, two questions. :)

.github/workflows/build-docs.yaml
Callisto13: there is also the option to do this:

> move the docs out of the profiles repo? or smth. they should be deployed whenever there are new releases in either pctl or profiles, so whatever makes it easier and more intuitive to do that idk

.github/workflows/deploy-docs.yaml
Callisto13: not sure i get what you mean.

do you mean:
- combine deploy and build actions

or
- merge the fetch bin and deploy steps within this action?

Skarlso: Yeah, I meant combine deploy and build. But maybe that would make that action too big?

.github/workflows/build-docs.yaml
Skarlso: Hmm. `profiles-docs`? That would make sense actually maybe. It would have a dependency on both, profiles and pctl. But at least would remove the dependency between them. ¯\\\_(ツ)\_/¯

.github/workflows/deploy-docs.yaml
Callisto13: deploy should only run on main merge, whereas the build should happen on PRs to check for broken links etc.

this used to be done implicitly as the `docs` make command was added to `test`, but i split it out for clarity

Skarlso: Ah okay, gotcha.

Callisto13: I don't actually know if the deploy action will work yet

yolo

Issue #214, opened 2021-07-05
As a profile author I want to be able to set dependency ordering between artifacts of the same kind


aclevername:
Prior spike https://github.com/weaveworks/profiles/issues/196

**What feature/behavior/change do you want?**
I want to be able to set dependency ordering between my charts or kustomizations. This is *ONLY* for artiacts of the same kind, e.g. helm chart depends on another helm chart, or kustomization depends on another kustomization. For a chart <-> kustomization relationship see #215.
**Why do you want this feature?**
I have a chart for deploying a CRD, and a chart that creates some instances of the CRD. I need the CRD to exist before I can create my instances, so I want to define that it should get deployed first.

## Proposed changes
Adding an additional `dependsOn` field that takes a list of artifact names:
```
spec:
  artifacts:
    - name: crd-chart
      chart:
        ...
    - name: instances-chart
      dependsOn: # new
      - name: crd-chart # new
      chart:
        ...
 ```
 
 When the HelmRelease/Kustomization resource gets generated for `instances-chart` it should contain the `dependsOn` fields. [Flux docs](https://fluxcd.io/docs/components/helm/helmreleases/#helmrelease-dependencies)

Example:
```
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: backend
  namespace: default
spec:
  interval: 5m
  chart:
    spec:
      chart: podinfo
      version: ">=4.0.0 <5.0.0"
      sourceRef:
        kind: HelmRepository
        name: podinfo
        namespace: default
      interval: 1m
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: frontend
  namespace: default
spec:
  interval: 5m
  chart:
    spec:
      chart: podinfo
      version: ">=4.0.0 <5.0.0"
      sourceRef:
        kind: HelmRepository
        name: podinfo
        namespace: default
      interval: 1m
  dependsOn:
    - name: backend
  ```

Skarlso:
Do we care if the profile does not exist? Flux will tell the user anyways. I think the build shouldn't care, just add the tag and move on IMHO. What do you think?

Skarlso:
I'm only asking, because there is a separation here, the builder doesn't have the list of all of the artifacts. It deals with a single one.

Skarlso:
[...]

PR #230, opened 2021-07-12
Adding dependsOn field to the artifacts

Skarlso:
### Description

Profiles part of https://github.com/weaveworks/profiles/issues/214

### Checklist


config/manager/kustomization.yaml
```diff
@@ -9,3 +9,9 @@ configMapGenerator:
 - files:
   - controller_manager_config.yaml
   name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: weaveworks/profiles-controller
+  newName: localhost:5000/profiles-controller
+  newTag: latest
```
Callisto13, line 17: is this that thing we said we would remember to not commit?

Callisto13 (approved):

config/manager/kustomization.yaml
Skarlso: Grrrr. We need to ignore this then force push it.

Skarlso:
@aclevername Yeah the force adding after ignoring doesn't work. :( Once it's tracked, it's tracked. No matter if it's ignored. We'll have to come up with something for this. :D

Skarlso:
We could run `git update-index --assume-unchanged [<file> ...]` but that's for local only.
Another way would be to somehow have a `.modified` file of this that we can ignore and that the tests are using.

Issue #231, opened 2021-07-12
Docs : change theme colours to match docs.gitops.weave.works


bia:
It would be a great start to match all of the theme colours (mostly links) and add the weaveworks logo (until we have our own logo) in order to make our docs site look cohesive with [docs.gitops.weave.works](https://docs.gitops.weave.works/docs/getting-started)

<img width="1440" alt="Screenshot 2021-07-12 at 15 58 05" src="https://user-images.githubusercontent.com/92439/125300031-f34ce800-e329-11eb-863e-d98b2d1dc570.png">

<img width="1440" alt="Screenshot 2021-07-12 at 15 58 54" src="https://user-images.githubusercontent.com/92439/125300177-12e41080-e32a-11eb-983c-ea14688fb5e1.png">

PR #233, opened 2021-07-15
Profiles dev documentation theme updates

### LINKED ISSUE #238: Docs branding: logos, pics, colourschemes
Profiles is an OS project which will be consumed by wego and other external products (ping in channel for more details).

We will need custom branding eventually. Currently in absence of any of that we are using the Weave logos and colourscheme, but this is not a permanent solution.

There are also lots of pics of dinosaurs hanging around (i.e. docusaurus branding), so they need to go too.

Himangini:
### Description

- [x] updated the docs to follow theme colours docs.gitops.weave.works
- [x] changed logos and pics from the default (remove all dinosaurs)
- [x] changed placeholder messages [still needs some design input on what these should be]
- [x] removed blog link from the header, added installation links instead
- [x] fixed site headers 
- [x] tweaked a few things
- [x] changed the alert components colors to match the weave ui components 

Screenshot 2021-07-19 at 20 13 09
Screenshot 2021-07-19 at 20 04 20

Fixes #238 , #231 
### Checklist


Callisto13:
Can we switch the order of `PCTL` and `Powered by Flux` on the homepage? The asymmetry of the texts hurt my delicate sensibilities 😂

userdocs/profiles.dev/docs/intro.md
```diff
@@ -18,7 +18,7 @@ Thank you!
 
 ## What are Profiles?
 
-In short, a profile is a Kubernetes package. Much as you would expect from another package
+In short, Profiles is a Kubernetes package manager. Much as you would expect from another package
```
Callisto13, line 21: technically, pctl is the package manager, and Profiles is a concept of package management

userdocs/profiles.dev/src/components/HomepageFeatures.js
```diff
@@ -4,29 +4,32 @@ import styles from './HomepageFeatures.module.css';
 
 const FeatureList = [
   {
-    title: 'GitOps Native package manager',
+    title: 'What are Profiles?',
     Svg: require('../../static/img/undraw_docusaurus_mountain.svg').default,
     description: (
       <>
-      some cool text here
+      Profiles is a GitOps Native Package Manager.
```
Callisto13, line 11: Yeh this is what i was struggling with around how we have named things. We don't have an actual "thing" called Profiles. Profiles is just the concept, and pctl is the actual tool which manages profiles (lowercase p) which are packages.

Callisto13: Also we have `What are Profiles?` (plural) then `Profiles is a..` (singular), which sort of illustrates what I am saying.

Profiles in the plural is more than one profile/package; `profiles are k8s packages` or whatever.

Profiles as a title in the singular is the concept: `Profiles is a GitOps package management mechanism` iyswim?

userdocs/profiles.dev/docs/intro.md
Himangini: yeah I know I was kinda struggling with it. Kinda tweaked it a little with your suggestion, imo this section should be related to `Profiles` since we are explaining `profile` in another section. wdyt looks better?

userdocs/profiles.dev/src/components/HomepageFeatures.js
Himangini: changed to make it consistent everywhere with your suggestion

userdocs/profiles.dev/docs/intro.md
Callisto13: yeh it's really tough, because people will read `Profiles` and immediately hear the plural of "more than one profile", so it will always be jarring to read `Profiles is...`.

If we go with the change you just added, logically it would make sense to change the heading to `What is Profiles?` which my brain just will not compute 😂 . This is why I originally had this bit talk about profiles (as in multiple units) rather than the concept.

So I think we have two options:
- On all user facing docs we don't use the word "Profiles" as a concept, just talk about "profiles" as an synonym for packages and how you use `pctl` to install them etc
- We keep using "Profiles" and "profiles", and in this bit here we remove the subheading (basically removing the `is/are` trap) and take care to avoid it in future.

Having written all that I am leaning to the former I think 🤔

Himangini: this seems to be going in circles, so I am going to put it back to what it was. Bianca suggested I should get people involved from product and/or marketing so we can come up with content for this, also we need some suggestions/content to showcase for the homepage. For now, I am adding few styling changes pointed out by Bianca and we can wrap this up :)

userdocs/profiles.dev/docs/intro.md
```diff
@@ -51,7 +51,7 @@ No worries, here's the breakdown:
 - A profile: an individual package of Kubernetes components. Lives in a git repo "upstream"
   of users who have installed it on their cluster
 - `N` profiles: more than one profile
-- Profiles (capital 'P'): a blanket term for this concept of GitOps native package management
+- Profiles (capital 'P'): a blanket term for this concept of GitOps native package management mechanism
```
Callisto13, line 54: this should either be:

> a blanket term for this concept of **a** GitOps native package management mechanism

or just 

> a blanket term for this concept of GitOps native package management

userdocs/profiles.dev/src/components/HomepageFeatures.js
```diff
@@ -4,41 +4,38 @@ import styles from './HomepageFeatures.module.css';
 
 const FeatureList = [
   {
-    title: 'GitOps Native package manager',
-    Svg: require('../../static/img/undraw_docusaurus_mountain.svg').default,
+    title: 'What are Profiles?',
```
Callisto13, line 7: The following line says "Profiles is ...", so this either needs to be:

> What is Profiles?

or

> What are 'Profiles'?

or neither

userdocs/profiles.dev/docs/intro.md
Himangini: fixed 👍🏻

userdocs/profiles.dev/src/components/HomepageFeatures.js
Himangini: fixed 👍🏻 , homepage is subject to change anyway

Callisto13 (approved):

PR #287, opened 2021-08-10
Bump deps

Callisto13:
Bump all the deps


aclevername (approved):

aclevername:
```
/home/runner/work/profiles/profiles/bin/controller-gen "crd:trivialVersions=true,preserveUnknownFields=false" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
../../../go/pkg/mod/k8s.io/client-go@v0.22.0/plugin/pkg/client/auth/exec/metrics.go:21:2: package io/fs is not in GOROOT (/opt/hostedtoolcache/go/1.15.14/x64/src/io/fs)
Error: not all generators ran successfully
```

Does one of our package bumps also require us to bump go versions? Or perhaps its that the Go version running on the ubuntu image for github actions is old

aclevername:
Looks like the pinning is only happening in the `Test` workflow, need to also add it to the `Acceptance` https://github.com/weaveworks/profiles/blob/9a5e9e271aed365ee1027e05d575d2c795fe0140/.github/workflows/test.yaml#L15-L17

Callisto13:
boooo

PR #293, opened 2021-08-11
use apiextensionsv1.JSON for default values field

### LINKED ISSUE #275: Use `apiextensionsv1.JSON` type for Helm default values
I had a look at the Helm related side of profiles, and was wondering if there is a specific reason the `defaultValues` (shown here: https://profiles.dev/docs/tutorial-basics/create-a-profile/) is of a type string.

Our experience with the helm-controller, because it actually represents a YAML structure, is to use the [`*apiextensionsv1.JSON` structure from the Kubernetes core](https://github.com/fluxcd/helm-controller/blob/v0.11.2/api/v2beta1/helmrelease_types.go#L150-L152). The added value of this is that it ensures things are properly validated (or can be validated pre-commit using tools like `kubeval`).

While the result is still passed on to a `ConfigMap` key/value (which doesn't really care about what you put into it, really), this at least ensures any major formatting issues are caught before the helm-controller starts borking.

aclevername:
### Description

Closes #275 

### Checklist


cmd/schema/schema_test.go
```diff
@@ -45,7 +45,7 @@ var _ = Describe("Schema", func() {
 		Expect(json.Unmarshal([]byte(contents), &s)).To(Succeed())
 		Expect(s.Version).To(Equal("http://json-schema.org/draft-07/schema#"))
 		Expect(s.Definition).ToNot(BeNil())
-		Expect(s.Definitions).To(HaveLen(9))
+		Expect(s.Definitions).To(HaveLen(10))
```
aclevername, line 48: @Callisto13 I don't quite get why this change of type caused the list to grow by 1... but I think its fine? :see_no_evil:

Callisto13: lol 🤷‍♀️

Callisto13: oh wait i think i know 🤔  can you update `cmd/schema/main.go` to remove the new thing whatever it is

PR #305, opened 2021-08-24
Don't generate readme when publishing docs

Callisto13:
Fixes the currently breaking docs deploy job


Callisto13:
hmmm some flakes in those tests

PR #363, opened 2021-11-16
Add ADR on layering.

bigkevmcd:
### Description

This is the ADR on layering.

### Checklist


docs/adrs/0003-profile-layering.md
```diff
@@ -0,0 +1,119 @@
...
+that some profiles should be installed before others if this is required.
+
+## Decision
+
+Support Helm Chart [annotation](https://helm.sh/docs/topics/charts/#the-chartyaml-file) that indicates a layer-like ordering for unrelated profiles.
+
+The `weave.works/layer` indicates that a Profile chart should be applied in a
+specific layer.
+
+For example, with the following two Profile charts.
```
Skarlso, line 35: Please provide a more elaborate configuration. Like 10 profiles at least so we can see how complex and convoluted this gets and how fast...

docs/adrs/0003-profile-layering.md
```diff
@@ -0,0 +1,119 @@
...
+      chart: demo-profile
+      version: "0.1.1"
+      sourceRef:
+        kind: HelmRepository
+        name: demo
+  dependsOn:
+    - name: base-profile-default
+```
+
+The layers are sorted lexicographically, which provides flexibility in naming,
```
Skarlso, line 94: I'm not following this one. Why should / would we sort layers?

docs/adrs/0003-profile-layering.md
```diff
@@ -0,0 +1,119 @@
...
+but we can determine a recommended layer naming strategy for co-ordinated use.
+
+## Profiles without layers
+
+Profiles without explicit layers should be configured to depend on the
```
Skarlso, line 99: Why? This is so that layering happens by default and we just say if they aren't configured than we just use "layer 0"?

docs/adrs/0003-profile-layering.md
bigkevmcd: The nice thing is that it doesn't get much more complicated...certainly simpler than explicit dependencies.

Skarlso (commented):
Looks okay, just have a couple questions to clarify.

docs/adrs/0003-profile-layering.md
Skarlso: For sure, but can we nest layers, and how would that look like in a bigger configuration?

Skarlso: I mean I guess nested layering is a noop since a has layer a and b has layer b and c can have layer b then since tha already has layer a.

docs/adrs/0003-profile-layering.md
bigkevmcd: We sort the layers...

If you are installing profiles with layers...
 * layer-0
 * layer-1
 * layer-2
This would be sorted, and so, profiles in layer-1 would depend on layer-0, and profiles in layer-2 would depend on layer-1.

Profiles in layer `""` (i.e with no layer) would end up depending on profiles in layer-2.

docs/adrs/0003-profile-layering.md
Skarlso: But what about multiple layers which aren't related? So C has layer a and layer b which do not co-depend?

docs/adrs/0003-profile-layering.md
bigkevmcd: Because we might want to have layers like `infrastructure` and `web` or something (there's no standard _yet_)

Skarlso: But what would sorting do in that regard? Sorry, I'm just not seeing where or what you mean by sorting here. :) Sort what, how and where? And where would the sorted result be at?

docs/adrs/0003-profile-layering.md
Skarlso: That's interesting. What if I don't care, and it should just be installed in parallel?

docs/adrs/0003-profile-layering.md
bigkevmcd: I think you're misunderstanding the layering, which means I need to do some work to clarify in the document.

The layers are for ordering of dependencies, and not for explicitly defining layers.

Skarlso: Oh. :)

docs/adrs/0003-profile-layering.md
bigkevmcd: I agree that it's somewhat arbitrary.

You could easily make an argument for any of the three options, "before everything else", "after everything else" and "at the same time as other things".

But it was a requested feature.

Skarlso: Gotcha. Okay. As long as we document it. :)

docs/adrs/0003-profile-layering.md
Skarlso: Ah, okay, it's clear now. The ordering of the dependencies is happening using the lexically sorted layer names. The feature is the name of the layer. :) Gotcha.

Skarlso (approved):
Ok, LGTM

Skarlso:
Can you do an update and go mod tidy on your branch please? :)

